PR #190, opened 2014-01-17
add freeaddrinfo to the hooks, fix a missmatch between glibc's addrinfo and bionic's addrinfo

krnlyng:
...


krnlyng:
was already merged

PR #251, opened 2015-01-13
dlsym the symbols in eglGetProcAddress from the glibc libraries

giucam:
eglGetprocAddress currently just returns the value it gets from the
android EGL if the procname isn't one of few special functions.
This has two issues:
1) Android's eglGetProcAddress returns garbage for core gl functions.
2) The calling convenction is different, which means float arguments
   to the returned functions will not work correctly.

To fix this get the symbols with the glibc's dlsym, using RTLD_DEFAULT
as library handle. If that fails we fall back to returning the value
of eglGetProcAddress, which may be sane.

### DIFF
--- a/hybris/egl/egl.c
+++ b/hybris/egl/egl.c
@@ -17,6 +17,8 @@
 
 /* EGL function pointers */
 #define EGL_EGLEXT_PROTOTYPES
+/* For RTLD_DEFAULT */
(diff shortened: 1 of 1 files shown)


thp:
Maybe limit the dlsym-based lookup to function names that start with "gl"? Other than that, LGTM.

PR #262, opened 2015-04-22
Prevent resizing problem on nexus 4

sledges:


hybris/egl/platforms/wayland/wayland_window.cpp
```diff
@@ -85,12 +85,18 @@ void WaylandNativeWindowBuffer::wlbuffer_from_native_handle(struct android_wlegl
...
+    lock();
+    this->m_defaultWidth = width;
+    this->m_defaultHeight = height;
+    unlock();
+}
+
 void WaylandNativeWindow::resize_callback(struct wl_egl_window *egl_window, void *)
 {
     TRACE("%dx%d",egl_window->width,egl_window->height);
-    native_window_set_buffers_dimensions(
```
giucam: Shouldn't this still be called in resize()?

PR #264, opened 2015-04-27
[libhybris] Don't leak fds of server allocated buffers

giucam:


giucam:
This actually does not do anything useful, the fds are closed when closing the native handle.

PR #267, opened 2015-04-28
[libhybris] Use CLOCK_REALTIME in combination with pthread waiting.

sletta:


stskeeps:
CLOCK_REALTIME is what's used internally in pthread_timedlock so this will do as API intended on bionic side.

morphis:
Can we add a comment above this in the source explaing why we choose CLOCK_REALTIME here? Don't want to end up with this discussion again.

PR #268, opened 2015-04-29
[libhybris] Make sure we're using the right clock types.

sletta:


sletta:
https://github.com/libhybris/libhybris/commit/8ee815023e3ca1f87b1d6631eceb360b84d12172 was preferred instead.

PR #269, opened 2015-06-02
[egl] eglGetProcAddress should return symbols of chosen EGL config

thp:
Previously we just looked up the process-global namespace for the functions, so looking up symbols only worked if the process also had the right `libGLES*.so` loaded.

EGL on other platforms doesn't need this, so make sure that we load the right library dynamically and return symbols from it. The library loaded is still the libhybris (glibc-based) one due to the hardfp calling convention.

This also avoids issues where `eglGetProcAddress()` would return any symbols in the global symbol namespace of the dynamic linker.

### DIFF
--- a/hybris/egl/egl.c
+++ b/hybris/egl/egl.c
@@ -40,6 +40,9 @@
 
 static void *_libegl = NULL;
 static void *_libgles = NULL;
+static void *_hybris_libgles1 = NULL;
+static void *_hybris_libgles2 = NULL;
+static int _egl_context_client_version = 1;
 
 static EGLint  (*_eglGetError)(void) = NULL;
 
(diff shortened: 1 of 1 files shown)


JussiPakkanen:
getenv("HYBRIS_LIBGLESV1") ?: "libGLESv1_CM.so"

Is the ternary operator guaranteed to return NULL here if getenv returns NULL? Even if not I'd recommending putting NULL there because being explicit is good.

thp:
The `?:` is a [GNU extension to C](http://en.wikipedia.org/wiki/%3F:#C):

```
a = x ? : y;
```
is equivalent to:
```
a = x ? x : y;
```
except that if x is an expression, it is evaluated only once.

So this means: "Either use the string in the environment variable `HYBRIS_LIBGLESV1` or if that is `NULL`, use the literal string `"libGLESv1_CM.so"`"

JussiPakkanen:
Fine by me. Don't have anything to say about what the code actually does, though.

thp:
Updated to use `"libGLESv1_CM.so.1"` and `"libGLESv2.so.2"`, as this is how the libraries are installed by libhybris.

thp:
I've now tested this on SbJ and am pretty confident that this improves the EGL `eglGetProcAddress()` situation.

PR #275, opened 2015-06-18
Unwrap EGLImage before passing it to android layer. JB#29670

tworaz:
Current implementation of EGL in libhybris wrapps EGLImage objects
received from android layer into a custom data structure. This means
that the object needs to be unwrapped properly before being passed back
to android EGL funcrion calls. This was done in hybris implementation of
libEGL, but not in libGLESv2. As a result whenever application (gecko in
this case) tried to use libGLESv2 version of EGLImageTargetTexture2DOES
it would fail with GL_INVALID_OPERATION.


thp:
LGTM, thanks for looking into this :)

PR #281, opened 2015-07-09
glesv1: parse DEFAULT_HYBRIS_LD_LIBRARY_NAME to find the correct (patched) version of libGLESv1_CM.so

krnlyng:
required for running glesv1 applications on sfos (jolla was not affected because there /system/lib/libGLESv1_CM.so is already patched) but on devices like nexus 5 this is required

don't know if it would be wiser to use getenv("LIBGLESV1") like the glesv2 wrapper does, please advice


krnlyng:
i have updated it to use a macro instead

krnlyng:
not necessary anymore :)

PR #288, opened 2015-08-31
Fix resolving the GLEsv2 functions at link time

giucam:
The issue was addressed in 295d159 but that commit had to be reverted by
c1099bd because of a crash when trying to dlopen the wrapped libGLESv2.so.
Instead, if a gl* function is resolved when we don't have yet a handle for
the library, use a wrapper calling the internal function.

### DIFF
--- a/hybris/glesv2/glesv2.c
+++ b/hybris/glesv2/glesv2.c
@@ -48,6 +48,133 @@ static void         (*_glVertexAttrib3f)(GLuint indx, GLfloat x, GLfloat y, GLfl
 static void         (*_glVertexAttrib4f)(GLuint indx, GLfloat x, GLfloat y, GLfloat z, GLfloat w) FP_ATTRIB = NULL;
 static void         (*_glEGLImageTargetTexture2DOES)(GLenum target, GLeglImageOES image) = NULL;
 
+static void         (*_glActiveTexture)(GLenum texture) = NULL;
+static void         (*_glAttachShader)(GLuint program, GLuint shader) = NULL;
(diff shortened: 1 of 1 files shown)


hybris/glesv2/glesv2.c
```diff
@@ -48,6 +48,133 @@ static void         (*_glVertexAttrib3f)(GLuint indx, GLfloat x, GLfloat y, GLfl
 static void         (*_glVertexAttrib4f)(GLuint indx, GLfloat x, GLfloat y, GLfloat z, GLfloat w) FP_ATTRIB = NULL;
 static void         (*_glEGLImageTargetTexture2DOES)(GLenum target, GLeglImageOES image) = NULL;
 
+static void         (*_glActiveTexture)(GLenum texture) = NULL;
+static void         (*_glAttachShader)(GLuint program, GLuint shader) = NULL;
+static void         (*_glBindAttribLocation)(GLuint program, GLuint index, const GLchar* name) = NULL;
```
vicamo: It seems original sym list is already sorted by name? So probably these proposed new symbols should be merged into the list instead.

morphis: Yeah, please order them correctly by name.

giucam: I kept them separate because they are needed for different reasons. The old ones are needed because they use floats, the new ones only because of the crash in android_dlsym(). If we find a way to fix that we can remove the new symbols and the wrapper funcs, and it's easier if they are separate.

morphis:
LGTM

PR #291, opened 2015-09-08
Drop compat layer

morphis:
Nobody else than Ubuntu uses this and it should be kept as out-of-tree
module using the provided APIs to wrap any other than the standard set
of Android libraries.


morphis:
LGTM from Stskeeps

PR #296, opened 2015-11-19
For upstream/fix android 6.0 compilation

vicamo:
This is only an initial work for Android 6.0 adoption. No functional verification is ever done.

Headers extracted with:
`${topsrcdir}/utils/extract-headers.sh `pwd`/android-6.0.0_r1 `pwd`/extracted/android-23 6 0 0`
Configured with:
`./hybris/autogen.sh --with-android-headers=../extracted/android-23 --prefix=/usr --with-default-egl-platform=hwcomposer --enable-debug --enable-trace --with-default-hybris-ld-library-path=/system/lib:/vendor/lib`

### DIFF
--- a/hybris/common/Makefile.am
+++ b/hybris/common/Makefile.am
@@ -10,11 +10,16 @@ if HAS_ANDROID_5_0_0
 SUBDIRS = jb
 libhybris_common_la_LIBADD= jb/libandroid-linker.la
 else
+if HAS_ANDROID_6_0_0
+SUBDIRS = jb
+libhybris_common_la_LIBADD= jb/libandroid-linker.la
+else
 if HAS_ANDROID_2_3_0
 SUBDIRS = gingerbread
 libhybris_common_la_LIBADD= gingerbread/libandroid-linker.la
 else
 $(error No Android Version is defined)
+endif
 endif
 endif
 endif
--- a/hybris/configure.ac
+++ b/hybris/configure.ac
@@ -178,6 +178,7 @@ AC_SUBST(ANDROID_VERSION_PATCH, [$android_headers_patch])
 AC_MSG_NOTICE("Android headers version is $android_headers_major.$android_headers_minor.$android_headers_patch")
 
 # Add automake tests for version/API needs here that you need in code, including test .am's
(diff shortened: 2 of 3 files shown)


morphis:
LGTM

vicamo:
Update the extractor script to include headers in _system/media/audio/include/system_. With a few more comparison, some header files are still missing in android-23:
```
Only in android-22/cutils: atomic-arm.h
Only in android-22/cutils: atomic-arm64.h
Only in android-22/cutils: atomic-inline.h
Only in android-22/cutils: atomic-mips.h
Only in android-22/cutils: atomic-mips64.h
Only in android-22/cutils: atomic-x86.h
Only in android-22/cutils: atomic-x86_64.h
Only in android-22/cutils: cpu_info.h
Only in android-22/cutils: dir_hash.h
Only in android-22/hardware: keymaster.h
```
The `cpu_info.h` was removed in [f21717c](https://android.googlesource.com/platform/system/core/+/f21717cd87a6b35822fe8bb17d4d28a76798a01e), `dir_hash.h` was removed in [d5ff3e9](https://android.googlesource.com/platform/system/core/+/d5ff3e9449a6b2ce388e09f86cd60f2231c033cd), `keymaster.h` was renamed to `keymaster0.h`, `keymaster1.h` and are still extracted.

For those atomic header files, they are removed and a new `stdatomic.h` was introduced in [bionic](https://android.googlesource.com/platform/bionic/+/android-6.0.0_r1/libc/include/stdatomic.h). I'm not sure if this file should be extracted, and where is it's new home.

stskeeps:
It seems a bit hazardous to bring in bionic atomics into glibc space (where these headers would be built).. where are they used?

vicamo:
@stskeeps Android libcutils have included `include/cutils/atomic.h` since AOSP was first released. In `system/core` it's mostly used in __libutils__ and more specifically the __RefBase__ class. Atomic functions never have been hooked before and after this patch.

To compile userspace program against Android `include/cutils/atomic.h` and glibc, by current patch, one will need `stdatomic.h` in his include paths. `stdatomic.h` is provided by C11 unless __\_\_STDC_NO_ATOMICS\_\___ is defined. See [cppreference](http://en.cppreference.com/w/c/atomic).

PR #297, opened 2015-11-22
open FB HAL before HWC, fixes HWC for some devices

mac-l1:
open FB HAL before HWC, fixes HWC for some devices (e.g. rockchip tvboxes): based on comment in surfaceflinger sources: https://android.googlesource.com/platform/frameworks/native/+/master/services/surfaceflinger/DisplayHardware/HWComposer.cpp#108

### DIFF
--- a/hybris/tests/test_hwcomposer.cpp
+++ b/hybris/tests/test_hwcomposer.cpp
@@ -119,7 +119,7 @@ void HWComposer::present(HWComposerNativeWindowBuffer *buffer)
 	assert(err == 0);
 
 	err = hwcdevice->set(hwcdevice, HWC_NUM_DISPLAY_TYPES, mlist);
-	assert(err == 0);
+	//assert(err == 0);
 	setFenceBufferFd(buffer, fblayer->releaseFenceFd);
 
 	if (oldretire != -1)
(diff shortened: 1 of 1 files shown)


hybris/tests/test_hwcomposer.cpp
```diff
@@ -119,7 +119,7 @@ void HWComposer::present(HWComposerNativeWindowBuffer *buffer)
 	assert(err == 0);
 
 	err = hwcdevice->set(hwcdevice, HWC_NUM_DISPLAY_TYPES, mlist);
-	assert(err == 0);
+	//assert(err == 0);
```
morphis: Why is this commented?

mac-l1: From code (4.2.2 & 4.4.2): basically because HWC in rockchip devices only supports HWC_DISPLAY_PRIMARY (=0), all other display types err; so setting all HWC_NUM_DISPLAY_TYPES (=3) by hwcdevice->set(hwcdevice, HWC_NUM_DISPLAY_TYPES, mlist) fails. And as a hack: hwcdevice->set(hwcdevice, 1, mlist) succeeds...

However, more importantly, comparing with android: in surfaceflinger (also in AOSP) this doesnt actually matter as the return value of mHwc->set() in https://android.googlesource.com/platform/frameworks/native/+/master/services/surfaceflinger/DisplayHardware/HWComposer.cpp#782 isnt used when called by https://android.googlesource.com/platform/frameworks/native/+/master/services/surfaceflinger/SurfaceFlinger.cpp#1255
However, here in test_hwcomposer.cpp the error is asserted and program stops.

morphis: That is fine. However you should drop the line completely (maybe replacing with a comment referring to surfaceflinger then).

mac-l1: Agree, just replaced the comment line.

mac-l1:
What more is needed to get this pull request merged?
After this i can pull another one implenting vsync as android does.

PR #300, opened 2015-11-23
utils: fix header extraction for ancient versions

vicamo:
* NFC headers is only available since v2.3 (API level 9)
* kernel headers was in bionic/ before v2.3 (API level 9)

Android API level tested: 4-10, 14-23

Signed-off-by: You-Sheng Yang <vicamo@gmail.com>

### DIFF
--- a/utils/extract-headers.sh
+++ b/utils/extract-headers.sh
@@ -203,9 +203,11 @@ elif [ $MAJOR -eq 5 ]; then
         system/core/libsync/include/sync
 fi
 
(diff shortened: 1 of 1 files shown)


vicamo:
The results can be found in https://git.launchpad.net/~vicamo/android-headers/+git/android-headers/log/?h=clean . Note that for Ubuntu Phone, we need `hardware/wifi.h` as well, so it's included in all API levels in my repository.

morphis:
LGTM

PR #307, opened 2015-12-16
Revert "use global hybris macros whenever available"

sledges:
With commit a6668d293c00b18b7cb755a0c7a600a56bc63339, on Sailfish OS lipstick (Wayland compositor) still works, but any clients (apps) crash upon start with ```Fatal: the server doesn't advertise the android_wlegl global!Aborted```

This PR reverts commit a6668d293c00b18b7cb755a0c7a600a56bc63339.

Thanks to @mlehtima for finding this

PR #308, opened 2015-12-17
use global hybris macros whenever available

vicamo:
This is a re-try after backed out in #307.


vicamo:
@sledges, this only differs from #307 in that `eglQueryString()` call. Do you mind having another test on it? Thank you.

sledges:
Tested on Nexus 5, no more crashes, thanks! :+1:

PR #316, opened 2016-06-20
[glesv2] Wrap glGetString() and return GL_VERSION == 2.0.

sletta:
libhybris doesn't expose any 3.0+ symbols so client code, like Qt,
becomes confused.

### DIFF
--- a/hybris/glesv2/glesv2.c
+++ b/hybris/glesv2/glesv2.c
@@ -21,6 +21,7 @@
 #include <dlfcn.h>
 #include <stddef.h>
 #include <stdlib.h>
+#include <stdio.h>
 
 #include <egl/ws.h>
 #include <hybris/common/binding.h>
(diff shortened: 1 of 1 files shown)


hybris/glesv2/glesv2.c
```diff
@@ -678,6 +679,13 @@ static void glGetShaderSource_wrapper(GLuint shader, GLsizei bufsize, GLsizei* l
 
 static const GLubyte* glGetString_wrapper(GLenum name)
```
morphis: Any reason why we can't merge this _wrapper into the actual function at the bottom?

Otherwise LGTM

sletta:
I was wondering the same. Since there was a wrapper to begin with, I thought it might have some purpose, so I put the code in the wrapper.

morphis:
Thanks

gerboland:
I hit this exact same issue:
https://bugs.launchpad.net/ubuntu/+source/qtubuntu/+bug/1594198
This patch will solve my issue too. Thanks!

morphis:
@sletta If you merge the wrapper into the actual function we can merge this.

morphis:
Thanks!

Issue #325, opened 2016-10-24
pthread counter


krnlyng:
the pthread counter implementation has one problem, there are a lot of functions which are called `*android*pthread*` which should be relatively safe to leave unhooked and there are internal functions in the blobs which also have pthread in their name, like `osal_cs_init(pthread_mutex_t*)` or `_Z12osal_cs_initP15pthread_mutex_t` when mangled.
- it is possible to add `|| strstr(symbol, "osal") != NULL` and `|| strstr(symbol, "android") != NULL` to https://github.com/libhybris/libhybris/blob/master/hybris/common/hooks.c#L1687
- or which i think is the better solution: replace the `strstr` in https://github.com/libhybris/libhybris/blob/master/hybris/common/hooks.c#L1684 with `strncmp(sym, "pthread", 7) == 0 || strncmp(sym, "__pthread", 9) == 0`

which solution should be picked or is there an even better one?

here is an example of a crash due to such a function:

> ...
> calling hooked method: pthread_mutex_lock@0xf7318561
> HOOKS hooks.c:1841 (get_hooked_symbol) DEBUG: _Z15osal_cs_destroyP15pthread_mutex_t -5
> 
> HOOKS hooks.c:1841 (get_hooked_symbol) DEBUG: _Z13osal_cs_enterP15pthread_mutex_t -6
> 
> HOOKS hooks.c:1841 (get_hooked_symbol) DEBUG: _Z13osal_cs_leaveP15pthread_mutex_t -7
> 
> HOOKS hooks.c:1841 (get_hooked_symbol) DEBUG: _Z12osal_cs_initP15pthread_mutex_t -8
> 
> calling unhooked method: __aeabi_atexit@0xf6f56a35
> calling hooked method: __cxa_atexit@0xf73b3375
> calling unhooked method: _ZN4llvm3sys9MutexImplC2Eb@0xf4a2597b
> calling unhooked method: __aeabi_atexit@0xf6f56a35
> calling hooked method: __cxa_atexit@0xf73b3375
> calling unhooked method: __aeabi_atexit@0xf6f56a35
> calling hooked method: __cxa_atexit@0xf73b3375
> calling unhooked method: _ZN4llvm3sys9MutexImplC2Eb@0xf4a2597b
> calling unhooked method: __aeabi_atexit@0xf6f56a35
> calling hooked method: __cxa_atexit@0xf73b3375
> calling unhooked method: __aeabi_atexit@0xf6f56a35
[...]

PR #326, opened 2016-10-27
don't hook all functions which contain pthread

krnlyng:
to negative numbers.

see #325


ghosalmartin:
I've tested on an arm device and it does indeed work :+1: :shipit:

PR #331, opened 2017-01-27
improvements

krnlyng:
some other improvements

### DIFF
--- a/hybris/common/mm/linker_debug.cpp
+++ b/hybris/common/mm/linker_debug.cpp
@@ -1,0 +1,686 @@
+/*
+ * Copyright (C) 2010 The Android Open Source Project
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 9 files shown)


morphis:
Can we split this in a few smaller PRs? LIke the https://github.com/libhybris/libhybris/pull/331/commits/4a92371334b0d28ade2c705766a95d661ec976f6 change can go into its own PR etc. 

Now that we've merged the whole ubuntu stuff lets come into a world were we do proper reviews and carefully look at individual changes to not break stuff.

krnlyng:
updated

hybris/common/mm/linker_debug.cpp
```diff
@@ -0,0 +1,686 @@
...
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+ * OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include "private/libc_logging.h" // Relative path so we can #include this .cpp file for testing.
+#include "private/ScopedPthreadMutexLocker.h"
+
+#include <android/set_abort_message.h>
```
morphis: set_abort_message.h is not part of what we fetch with our extra headers script. This is also why CI is failing. So lets import this header file into the mm linker copy we maintain here.

morphis:
Left a comment about failing CI inline. Can you describe more what this PR is supposed to do? Just "improvements" is not saying anything.

krnlyng:
it is basically work in progress and maybe also needs to be split uo, i left it here in the original pr because i wasn't sure about it yet. it readds the gnu indirect functions which needs testing on devices where that caused problems, it uses __thread (which i hope does not cause problems) to store the dlerror strings instead of a hack, it allows building with std=c++0x since some build environments lack updated gcc versions, it reintroduces the debug env variables as we talked about and it renames some functions which are also in glibc to avoid symbol name clashes which might cause problems (but seems to not have so far) all of this seems optional to me at this point.

krnlyng:
all the important fixes where moved to other PRs, this was just a left over. i will look if there is still anything useful in this

PR #334, opened 2017-01-27
hybris: fix mismatch between bionic fpos_t size and glibc fpos_t in bionic_file struct

krnlyng:
to avoid a crash on exit of most programs


morphis:
LGTM

PR #335, opened 2017-02-04
linker: introduce HYBRIS_LINKER_FORCE_BASENAME for .so search

Tofee:
If HYBRIS_LINKER_FORCE_BASENAME is set, then the linker will first
look for the .so based on the basename of the library, in case an
absolute path has been given.
This way it is possible to redirect .so loading even for absolute paths.


morphis:
LGTM

PR #337, opened 2017-02-16
hooks: include mm hooks only when WANT_LINKER_MM is defined

Tofee:


morphis:
LGTM

PR #338, opened 2017-02-16
hooks: remove unneeded TLS variable

Tofee:
The two "buf" variables don't need to be stored in TLS or to be global,
as they are only used as intermediate variables for the final glibc call.
Therefore remove the "static __thread" prefixes for these two.

Note: In my case having these two TLS variables did also lead to crashes when the
"tls_hooks" TLS variable is returned in the get_tls_hooks function. However
I couldn't determine conclusively why it crashed in that case.


morphis:
LGTM

PR #342, opened 2017-04-01
Fix the makefile generated by the extract-headers.sh script

bhush9:
Earlier it didn't install the log directory, which resulted in failure
if headers are installed to system-wide location


bhush9:
bump!

morphis:
LGTM

PR #344, opened 2017-04-20
(WIP) Enable initialize bionic

krnlyng:
This adds an optional flag to the automake based build: --enable-initialize-bionic

It enables initialisation of bionic libc which was disabled since it breaks. Although with https://github.com/krnlyng/android_bionic/commit/dd76989317a72896960a4162e3010102e4b96143 as well as https://github.com/krnlyng/android_bionic/commit/aaf4ebae112028f6b02cb2c64123e914d0422111 and https://github.com/krnlyng/droidmedia/commit/6de4887d0149464679b4cf88673fea130bccc8a0 this does work properly.

The bionic patch allows us to initialize bionic's libc by not initializing some TLS related things inside bionic libc. Functions which use TLS still need to be hooked. The purpose of the droidmedia patch is to have working camera in both cases, when bionic libc is initialized and when not.

The purpose of this patch is to have minimal amount of hooks and the purpose of having a minimal amount of hooks is that we don't need as much wrappers for bionic functions and  the android blobs can just call most of the bionic libc functions without us having to worry about different side effects of using the wrong implementation of those libc functions.

This means stuff does break LESS often, it has enabled us to fix many of the remaining issues with cm13 based ports. Inexplicable crashes become a matter of debugging only a handful of functions, namely pthread and a few others.

One downside is that we need the droidmedia patch. Since in minimal hooks we don't hook *alloc/free functions (since it has been shown to break stuff) we need this workaround.
The workaround adds a function glibc_malloc to bionic which can be called from bionic space. This allocated memory can then be passed to glibc space and free'd there with glibc's free method. The default behaviour of glibc_malloc in both (minimal hooks, non-minimal hooks) case is to call glibc's malloc (either directly or via the malloc hook). Which means this should be non-intrusive to existing ports. If someone does not want to use any of this --enable-initialize-bionic can be ommitted and the other two patches not applied.

Please comment on this WIP.


morphis:
This is pretty nice! Would like to see this tested on different platforms to see how it works there. Otherwise keeping it behind a configure option is fine for the moment. Tell me once you're happy with this and it isn't WIP anymore.

krnlyng:
hi, a quick update.

while this resolves a few issues there are new problems:

ideally we would hook *alloc & free methods in this too since then we wouldn't need the droidmedia patch, and to make it absolutely safe we would redirect bionic libc internal *alloc, free calls also to our wrappers (via a jumptable) to avoid glibc_free(bionic_malloc(s)) cases.

but that causes weird crashes and shader compilation errors.

the thing is if i allocate more memory than required _hybris_hook_malloc(size_t s) { return malloc(PAGE_CEILING(s)); } the crashes/errors do not occur, could this be a bug in the blobs?
also using LD_PRELOAD=/path/to/compiled/libjemalloc.so app does avoid the crashes.

eliminating this issue would make it non-WIP.

krnlyng:
the glibc_malloc patch is not necessary anymore, but i have had to add a flag --enable-adreno-quirks in order to allocate 5 bytes instead of 4 if malloc(4) is called. this causes a slight overhead obviously but resolves the problem.

so in order to have this working only these libhybris patches and https://github.com/krnlyng/android_bionic/commit/aaf4ebae112028f6b02cb2c64123e914d0422111 are required in order to use --enable-initialize-bionic (--enable-adreno-quirks)

another option in order to avoid --enable-adreno-quirks is to use jemalloc in glibc environments.

krnlyng:
Since the adreno quirk is not ideal i also tried the following (there is actually a memcpy to the address which was previously allocated with 4 bytes):

```
static void *_hybris_hook_memcpy(void *dst, const void *src, size_t len)
{   
    TRACE_HOOK("dst %p src %p len %zu", dst, src, len);

    if (src == NULL || dst == NULL)
        return dst;

#ifdef WANT_ADRENO_QUIRKS
    if(len == 5 && malloc_usable_size(dst) < 5) {
        return memcpy(dst, src, 4);
    }
#endif

    return memcpy(dst, src, len);
}
```
but still there are random segfaults and shader compilation errors, i currently don't see another way to fix this other than increasing the malloc(4) to malloc(5).

krnlyng:
should this be split up into two PRs? --enable-initialize-bionic is not required anymore on those devices where it was required: since the addition of --enable-adreno-quirks

we have a lot of quirks now (e.g. --enable-adreno-quirks --enable-mali-quirks and https://github.com/libhybris/libhybris/issues/352) for existing problems with different blobs. how should we properly handle those? i can investigate more whether we're doing something wrong which causes the bugs in the first place but it really seems like bugs in the blobs.

sledges:
> i could split it up like this (in order of importance):
> 1. --enable-adreno-quirks

This split would be more than welcome.

> we have a lot of quirks now (e.g. --enable-adreno-quirks --enable-mali-quirks and #352) for existing problems with different blobs. how should we properly handle those?
E.g. --enable-adreno-quirks started to manifest itself in 64bit devices, but we can't really stamp it on all of them; I'd go with empirical approach (and indicate per-device per-packaging) until a some sort of autodetection could be carried out (if ever due to this being in the blobs...)

krnlyng:
will split it up and resolve the conflicts :)

morphis:
> will split it up and resolve the conflicts :)

Sounds like recent activity to merge outstanding PR brought your attention back :-)

PR #347, opened 2017-05-05
sf: 2 new functions and test_sf

mickybart:
Reintroduce test_sf (previously dropped and partially "reverted" later)

Add 2 new functions to sf that are independent and will not break existing code :
- sf_get_display_info
- sf_client_get_egl_context

sf_get_display_info :
permit to get more details that are not necessary available with current code that requesting the framebuffer. We directly do the request to SurfaceFlinger.

sf_client_get_egl_context : 
permit to get the EGL context of a SfClient. This is useful for QPA.


ghosalmartin:
these changes work great for me on the Huawei P9 since it has no hwcomposer support

morphis:
LGTM

PR #348, opened 2017-05-11
hybris (arm): introduce HYBRIS_TRACE_(DYNHOOKED/UNHOOKED/HOOKED)

krnlyng:
useful to trace calls to and between android blobs.

shows you (somewhat) what the blobs are actually doing

works for functions which have no debug hook (HOOK_DIRECT_NO_DEBUG) but cannot print arguments, nevertheless i found it useful many times

implemented by injecting a wrapper function between a function that is called and it's actual implementation

it is entirely optional (--enable-mm-arm-tracing)

for example you could run
HYBRIS_TRACE_UNHOOKED=1 test_egl
and IF it would say: calling unhooked method malloc@ptr
you would know that malloc is unhooked (which it shouldn't be) and is called.

if for example test_egl crashes you could run
HYBRIS_TRACE_HOOKED=1 test_egl
and if you're lucky it would tell you maybe: calling hooked method foo@ptr with a crash immediately afterwards -> it's likely that there is a bug in this methods wrapper.

HYBRIS_TRACE_DYNHOOKED=1 would show you which functions are called which where previously loaded via dlsym and is a nice way of tracing egl/gl calls.

limitations:
* currently only implemented for the mm linker (would be possible to port)
* currently only available for arm 32 bit
* cannot show you which symbol is referenced, only which functions are called

### DIFF
--- a/hybris/common/Makefile.am
+++ b/hybris/common/Makefile.am
@@ -18,11 +18,27 @@ libhybris_common_la_SOURCES = \
 	strlcat.c \
 	logging.c \
 	sysconf.c
+
+if WANT_ARM_TRACING
+libhybris_common_la_SOURCES += \
+	wrappers.c \
+	wrapper_code_generic_arm.c
+endif
+
 libhybris_common_la_CPPFLAGS = \
 	-I$(top_srcdir)/include \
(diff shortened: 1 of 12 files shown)


morphis:
LGTM

PR #353, opened 2017-07-12
[test_vibrator] Add support for new Android vibrator API.

mlehtima:
Needed on Android 7 based devices because legacy header was removed.


morphis:
LGTM

PR #354, opened 2017-09-04
hybris: introduce --enable-adreno-quirks

krnlyng:
If this is enabled malloc(4) will be treated as if it was malloc(5).

This introduces overhead obviously but solves the following problem:

If sailfishos-browser crashes for no reason with shader compilation
errors this is required. It is most likely a bug in the blobs since
we only experience it on a subset of devices. Also using a jemalloc
implementation as LD_PRELOAD to libhybris works around this problem.
jemalloc is used in android which seems to indicate that it is more
lenient with these kinds of out-of-bounds accesses.

NOTE: sailfishos-browser is not the only affected app, it is just
an example. lipstick and others are also affected but it is not as
noticeable. Reboot/UI-restart is necessary after this change.


morphis:
LGTM

PR #355, opened 2017-09-08
dlerror() should return NULL on success

krnlyng:
some code depends on dlerror() returning NULL when dlsym succeeds.


morphis:
LGTM

PR #356, opened 2017-09-23
Sync with Halium 7.1

mariogrip:


mlehtima:
The mm linker won't work with all android 7 binaries and should not be use by default, a proper android 7 linker for libhybris exists for sailfish and we'll try to get it included in upstream hopefully soon.

morphis:
LGTM

PR #360, opened 2017-10-17
N linker + fixes

krnlyng:
This is the nougat linker for libhybris.
Some hwcomposer implementations (and other libraries) use the new android libdl namespaces, therefore we need a new linker for newer ports.


krnlyng:
I have added two more patches from @sledges which makes it easier to detect which linker to build and which quirks to use. Also the mm linker is not considered experimental anymore but simply enabled when the android version is 6 or higher.

sledges:
Those two patches are WIP, do not merge this in yet.

krnlyng:
Tested extensively, removing WIP status. Let me know if changes are necessary.

sledges:
LGTM from me

krnlyng:
adding a commit to adapt to the upcoming /odm partition paths. i.e. allow libraries to be loaded from /odm.

morphis:
LGTM

PR #363, opened 2017-11-10
Assign the layer list only to the first display

NotKit:
It's done the same way in https://github.com/mer-hybris/qt5-qpa-hwcomposer-plugin/blob/master/hwcomposer/hwcomposer_backend_v11.cpp#L229

This fixes test_hwcomposer on MediaTek Helio X10 device (Xiaomi Redmi Note 2), as otherwise HWC will freeze if there are no other displays connected.

### DIFF
--- a/hybris/tests/test_hwcomposer.cpp
+++ b/hybris/tests/test_hwcomposer.cpp
@@ -218,7 +218,11 @@ int main(int argc, char **argv)
 
 	int counter = 0;
 	for (; counter < HWC_NUM_DISPLAY_TYPES; counter++)
-		mList[counter] = list;
+		mList[counter] = NULL;
+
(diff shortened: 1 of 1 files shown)


z3ntu:
Coding style is off

NotKit:
@z3ntu fixed, thanks for noticing.

IlyaBizyaev:
This commit helps solve issues with Intel's HWComposer as well (half of https://github.com/libhybris/libhybris/issues/368).

IlyaBizyaev:
KWin does the same way: https://cgit.kde.org/kwin.git/tree/plugins/platforms/hwcomposer/hwcomposer_backend.cpp

morphis:
LGTM

PR #367, opened 2018-01-03
fbdev_window: wait for fence to be signalled and close it

kszaq:
If a sync fence is provided in queueBuffer, we have to wait for it to signal (for vsync) and close it (so that fd's are not leaking).


Ntemis:
why this didn't merge yet? any issues to be resolved?

Ntemis:
@morphis pong for merge

Issue #357, opened 2017-10-08
mm.so segfault


IlyaBizyaev:
Hello! :)

I am porting Halium to ASUS ZenFone 5 (A501CG, x86) and facing this issue when running `test_egl_configs`:

```
Starting test (EGL_PLATFORM=(null))
WARNING: linker /system/vendor/lib/egl/libEGL_POWERVR_ROGUE.so: unused DT entry: type 0x1d arg 0x559
WARNING: linker /system/vendor/lib/libIMGegl.so: unused DT entry: type 0x1d arg 0x7cc
WARNING: linker /system/vendor/lib/libsrv_um.so: unused DT entry: type 0x1d arg 0xbcc
Segmentation fault
```

GDB stack trace:
```
Program received signal SIGSEGV, Segmentation fault.
0xb7ba2840 in soinfo::call_constructors (this=0xb77ca23c) at linker.cpp:2228
2228    linker.cpp: No such file or directory.
(gdb) bt                                                                                     
#0  0xb7ba2840 in soinfo::call_constructors (this=0xb77ca23c) at linker.cpp:2228             
#1  0xb7ba62b7 in do_dlopen (name=0x8057898 "/vendor/lib/egl/libEGL_POWERVR_ROGUE.so", flags=0, extinfo=0x0) at linker.cpp:1634
#2  0xb7b9f508 in dlopen_ext (filename=0x8057898 "/vendor/lib/egl/libEGL_POWERVR_ROGUE.so", flags=0, extinfo=0x0) at dlfcn.cpp:95
#3  0xb7dc4672 in _hybris_hook_dlopen (filename=0x8057898 "/vendor/lib/egl/libEGL_POWERVR_ROGUE.so", flag=0) at hooks.c:2401
#4  0xb7b8718b in ?? ()                                                                      
#5  0xb7b86adc in ?? ()                                                                      
#6  0xb7b80514 in ?? ()                                                                      
#7  0xb7b807ca in ?? ()                                                                      
#8  0xb7fcae2b in eglGetDisplay (display_id=0x0) at egl.c:193
#9  0x0804874b in main (argc=1, argv=0xbffff7f4) at test_egl_configs.c:96
```

strace:
```
execve("/usr/bin/test_egl_configs", ["test_egl_configs"], [/* 15 vars */]) = 0
brk(NULL)                               = 0x80f6000
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
[...]

PR #369, opened 2018-01-14
Fix nullpointer crashes in strcmp

IlyaBizyaev:
Fixes this issue: https://github.com/libhybris/libhybris/issues/357
Credits go to @ruleh

### DIFF
--- a/hybris/common/mm/linker.cpp
+++ b/hybris/common/mm/linker.cpp
@@ -2227,11 +2227,7 @@ void soinfo::call_constructors() {
     return;
   }
 
-  if (soname_ == nullptr) {
-    soname_ = basename(realpath_.c_str());
-  }
-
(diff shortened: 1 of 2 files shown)


krnlyng:
I am wondering if the linker could possibly get confused by this if DT_SONAME is not set in the library but soname_ gets set by this change. Since some parts of the n/mm code are prepared for get_soname() == nullptr. I think that needs to be checked, in order to avoid possible regressions.

If that cannot possibly cause any problems, then LGTM.
If it can cause problems then the proper solution would be to wrap the hybris-libc.so check into an if(soname_ != nullptr)

IlyaBizyaev:
You are absolutely right, that change could affect other code, and I am sorry for that. I modified the check to be nullptr-safe, that seems to be the best solution.

morphis:
LGTM

PR #370, opened 2018-01-14
Fix TLS slots to be compatible with x86

IlyaBizyaev:
In a 4 year old commit, a Canonical employee "shifted TLS slots to avoid conflicts with libc/hybris": https://code-review.phablet.ubuntu.com/gitweb?p=aosp/platform/bionic.git;a=commit;h=7b1fefd038b943944be08219270ea1acc31b6b3d

While the change seems to work with ARM devices, it is a no-go for x86 ones. In fact, that patch changes `TLS_SLOT_STACK_GUARD` slot value while the comment tells specifically not to do that.

The change (due to some coincidence) did not harm up to Android 5.0. In more recent versions, applications compiled with those slot values crash immediately due to stack corruption.

All the x86 porters I know who work with recent Android versions modify TLS slots locally. So, I suggest to upstream these tested values that return `TLS_SLOT_STACK_GUARD` to its place, shifting other slots around.

### DIFF
--- a/hybris/common/mm/bionic/libc/private/bionic_tls.h
+++ b/hybris/common/mm/bionic/libc/private/bionic_tls.h
@@ -51,12 +51,17 @@ __BEGIN_DECLS
 enum {
   TLS_SLOT_SELF = 0, // The kernel requires this specific slot for x86.
   TLS_SLOT_THREAD_ID,
-  TLS_SLOT_ERRNO = 5,
+
+  TLS_SLOT_STACK_GUARD = 5, // GCC requires this specific slot for x86.
+
+  TLS_SLOT_ERRNO = 6,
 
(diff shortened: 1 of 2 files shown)


krnlyng:
These defines are not used by libhybris, in fact all uses of them which come from bionic are wrapped inside: #ifdef DISABLED_FOR_HYBRIS_SUPPORT.

So the changes would neither harm, nor fix anything. Those need to be applied to android_bionic (and i've seen you did that).
IMHO, the best approach would be to remove those bionic tls headers entirely, in order to avoid further confusion.

IlyaBizyaev:
Yeah, sure, if they are not needed, removal is better :)

IlyaBizyaev:
This change is indeed not needed, waiting for android_bionic patch to be merged.

PR #371, opened 2018-01-17
eglplatform_fbdev: open new framebuffer when creating a window

kszaq:
Framebuffer device has to be reopened when creating a new window to adapt its size to display. This is required e.g. after screen resolution changed.

Without reopening framebuffer, its dimensions remain from before resolution change.

This was tested with LibreELEC on Amlogic S912 device using libhybris for EGL.

### DIFF
--- a/hybris/egl/platforms/fbdev/eglplatform_fbdev.cpp
+++ b/hybris/egl/platforms/fbdev/eglplatform_fbdev.cpp
@@ -29,13 +29,6 @@ extern "C" void fbdevws_init_module(struct ws_egl_interface *egl_iface)
 		assert(0);
 	}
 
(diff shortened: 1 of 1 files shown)


morphis:
I am not sure this applies to all platforms as they have different semantics about when the framebuffer should be opened as far as I remember. Can anyone give this a try on other platforms?

krnlyng:
@kszaq on many platforms framebuffer needs to be opened before gralloc, this change would break those platforms. Can this be changed to open the framebuffer as originally and only close and reopen the framebuffer when it is required, if it is possible to detect whether reopening is required that would be nice (i.e. detect the resolution change), but if that is not possible maybe this behaviour could be enabled via an environment variable.

PR #374, opened 2018-01-29
Add screen rotation option for devices with upside down screen. Envir…

Michal-Szczepaniak:
…onment variable HAL_TRANSFORM_ROT


bhush9:
PS: commit message and code have two different env var, maybe fix one :)

Michal-Szczepaniak:
Right… fixed :P

morphis:
LGTM

PR #375, opened 2018-01-29
Look for libhardware.so from default search path.

NotKit:
This restores commit 98a8779f2235b9a57f9c1d16bf1e77f3c96f267c, which was lost during the merge of Ubuntu version.

It's needed for arm64 libhybris to work (where libhardware is located at /system/lib64), as well as for some SailfishOS devices which require patched libhardware.so.


morphis:
LGTM

PR #377, opened 2018-03-09
Fix splitting if there is more than one separator

Michal-Szczepaniak:
When there is more than two delimiters its searching and splitting by whole string not by each character in string.That fixes it.

### DIFF
--- a/hybris/common/n/linker.cpp
+++ b/hybris/common/n/linker.cpp
@@ -210,7 +210,7 @@ static const char* const kAsanDefaultLdPaths[] = {
 std::vector<std::string> split(const std::string &text, const std::string &sep) {
   std::vector<std::string> tokens;
   std::size_t start = 0, end = 0;
-  while ((end = text.find(sep, start)) != std::string::npos) {
+  while ((end = text.find_first_of(sep, start)) != std::string::npos) {
     tokens.push_back(text.substr(start, end - start));
     start = end + 1;
   }


krnlyng:
just one question: LD_SHIM_LIBS is supposed to be set only once to the same value for all binaries in android. do we need some way of determining that LD_SHIM_LIBS variable automatically instead of reading it from that value? if we do it at runtime (parsing init.environ.rc) that will surely lead to permission erros) so maybe we could do it differntly, i.e. having a DEFAULT_HYBRIS_LD_SHIM_LIBS variable (similar to DEFAULT_HYBRIS_LD_LIBRARY_PATH) that would be determined by the hybris build system and passed to autoconf/configure. that way we could have it set automatically to the proper value. and we could still also read that env variable such that one-build-rules-all-the-devices will not be affected by this. (i.e.

having that autoconf variable would allow us to avoid thinking about that env variable (i.e. we would not need to set it per-process, which would be kind of lame anyhow since that variable is usually set just once for all processes?)

IlyaBizyaev:
Thank you for your patch, it seems to work for me!

Unfortunately, at first it failed to build. There's a declaration:
```
const char* ldshim_libs_env = nullptr;
```
missing in L4636. Could you add it? :)

Michal-Szczepaniak:
I thought i added it. Right away!

krnlyng:
re my comment, seems in android they also did that (but not for convenience, but for security): https://github.com/LineageOS/android_bionic/commit/a6fb1155c030661d320d895ae39b78211f57a646#diff-3175b426046787eece7377387340b982

IlyaBizyaev:
The problem I can see with compile-time approach is that it won't be usable in Halium and distributions based on it (e.g. Plasma Mobile).

krnlyng:
@IlyaBizyaev yes that's what i also considered in my comment and that shouldn't be an issue. though it is an issue with future android versions where this is built in by default.

IlyaBizyaev:
Oh, I see... Yeah, that sounds right

krnlyng:
@Michal-Szczepaniak should we instead read this variable from a configuration file?

Michal-Szczepaniak:
@krnlyng im not sure (im not that advanced) but its exported in init.* file https://github.com/OpenKirin/android_device_huawei_alice/blob/7.x/rootdir/etc/init.hi6210sft.rc#L165

krnlyng:
@Michal-Szczepaniak yes but most likely we cannot read that file, since processes using libhybris generally don't have high enough privileges

Michal-Szczepaniak:
@krnlyng but since its set in android in one file then configuration file should be okay as its set once and doesnt break anything

krnlyng:
@Michal-Szczepaniak I reconsidered and it's probably ok to merge though it's not ideal.

But can you apply this patch to all of the linkers and not just n?

Michal-Szczepaniak:
absolutely (even if i dont need it anymore :p)

Michal-Szczepaniak:
@krnlyng Hmm parsing shims adds them to g_ld_all_shim_libs but i don't see this variable being used anywhere, so if i understand everything correctly this change doesn't work at all.

JCLemme:
This change is necessary for the Amazon Fire 7 (austin) port.

PR #390, opened 2018-06-18
hybris: Use correct struct in glEGLImageTargetTexture2DOES of glesv2

rainemak:
See also egl.c


rainemak:
@krnlyng @sledges review ?

krnlyng:
LGTM

krnlyng:
one moment: in GLESv1_CM this is not yet implemented. there are not many users of GLESv1_CM but we should fix it there too.

rainemak:
Updated glesv1_cm.c

PR #393, opened 2018-07-05
linker N: Improve gdb support

Tofee:
With default bionic's implementation, the list of loaded libraries
for GDB is reset when the first Android lib is loaded.
With libhybris, we want to get both Android and Glibc loaded libraries.

However, we shouldn't simply link our "link_map" pointers in the
current list, because glibc's "link_map" object shouldn't point to
our incompatible "link_map" objects. To avoid that:
 1. make a copy of the initial list, to avoid modifying glibc's objects
 2. prepend our objects to the list, instead of appending it

 That way glibc's objects are never modified, and glibc's linker won't
 try to manipulate objects that haven't been initialized by him.


krnlyng:
@Tofee should we consider adding this to other linkers as well or are they already good?

Tofee:
Other linkers do it differently, iirc. But I didn't look at them very carefully.

krnlyng:
@Tofee ok, i'll take a note to check it.

PR #397, opened 2018-08-29
egl: Fix building nativewindowbase with android-headers-24

beidl:
The sync/sync.h header does not exist from Android 7.0 onwards, hence use the
newer buffer fencing methods instead of the deprecated ones starting ANDROID_VERSION_MAJOR >= 7.


beidl:
WIP pull request, still testing.

PR #398, opened 2018-09-02
Add NFC wrapper generator scripts

thp:
Since the external `libhybris-nfc-wrapper-generator` repository does not exist anymore, I've dug out the old code and created a PR so that the wrapper scripts live in the same repos as libhybris proper.

The original history of the other repo is included, I decided to remove the submodule references for the Android source code to not unnecessarily bloat checkouts, checkout commit 36f0399 for the vanilla state of `libhybris-nfc-wrapper-generator`.

### DIFF
--- a/hybris/libnfc_ndef_nxp/libnfc_ndef_nxp.c
+++ b/hybris/libnfc_ndef_nxp/libnfc_ndef_nxp.c
@@ -17,8 +17,7 @@
  */
 
 /**
- * Bindings generated using:
- * https://github.com/thp/libhybris-nfc-wrapper-generator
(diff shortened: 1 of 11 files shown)


krnlyng:
In mer-hybris builds we usually have libhybris checked out in $ANDROID_ROOT/hybris/mw/libhybris
(nowadays sometimes also in $ANDROID_ROOT/external/libhybris but this location might be disappearing eventually). So we could change utils/generate_nfc/README.android-src to reflect these paths and point the utils/generate_nfc/generate.sh script to those paths as well.

Thanks, i think this looks good, i will take care of it soon.

PR #404, opened 2019-01-10
hooks: handle pthread_cleanup_push and _pop

jusa:


bhush9:
Just curiosity, was this causing any issues without it?

jusa:
> Just curiosity, was this causing any issues without it?

Not in current adaptations afaik but I came across one audio adaptation which would segfault without the hooking.

Issue #402, opened 2018-12-19
Crashing issue in eglplatform_wayland on m_window in a race condition


dcaliste:
I'm facing seldom crashes of a Wayland client (namely lipstick-security-ui). Backtrace is pointing in eglplatform_wayland.so, more precisely in `WaylandNativeWindow::dequeueBuffer#376`. There m_window is NULL on crash. I've added debug messages in this routine and in free_callback. During normal behaviour, I've got something like:

<pre>
We are dequeuing here for what reason, don't know 0x1e626d0 (0x1e5da98).
We are dequeuing here for what reason, don't know 0x1e626d0 (0x1e5da98).
We are dequeuing here for what reason, don't know 0x1e626d0 (0x1e5da98).
Ok, free callback called, no more window here 0x1e626d0 (0x1e5da98).
</pre>
The first pointer address is the value of m_window, while the pointer in parenthesis is the value of `this`.

When the crash happens, it appears that the free_callback routine has been called before the dequeueing routine:
<pre>
We are dequeuing here for what reason, don't know 0x2004468 (0xb050c600).
We are dequeuing here for what reason, don't know 0x2004468 (0xb050c600).
Ok, free callback called, no more window here 0x2004468 (0xb050c600).
We are dequeuing here for what reason, don't know (nil) (0xb050c600).
</pre>

I'm discovering the Wayland protocol and cannot comment much more on why the dequeueing operation is happening here after the free_callback. The user situation is the dismiss of a system window. My guess is that, the `finishSwap()` is generating a frame_callback that is waiting for events (thus the dequeue) while the main widow (m_window here?) has been already destroyed.

Any idea how to treat this (besides returning from the dequeue if m_window is NULL already). In addition, there are other places where m_window is used without checking that it is not NULL, so maybe it's something that should be think about also.

(@jpetrell and @chriadam if you are interested)

PR #405, opened 2019-01-16
Fix crashes in threaded EGL wayland rendering

dcaliste:
Destroying an EGLSurface object in the main thread while a rendering thread is calling eglSwapBuffers() results in various crash / hanging scenarios:
* a segfault in dequeueBuffer() while calling m_window->width();
* a hang in finishSwap() because readQueue() is called blocking on the frame_callback, while this ressource have been killed already and there will be no event in the queue;
* a segfault in finishSwap() while creating the frame_callback with m_window->surface.

This PR is adding more protection against illegal use of m_window, even if the case described above is an illegal case (calling destroying a window while the rendering thread is swapping it). It should solve #402.

What do you think?


chriadam:
@adenexter @krnlyng this one is related to https://bitbucket.org/snippets/dcaliste/LeaArX/crashes-with-lipstick-ui-security

Issue #389, opened 2018-06-11
Would it be ok to split some things?


krnlyng:
As per: https://github.com/mer-hybris/droid-hal-device/pull/185#issuecomment-396295730

Would it be ok to split libhybris/compat off into a seperate libhybris-compat (or even hybris-compat) repository?
Only if it doesn't introduce any big problems. (Changing a manifest.xml to a different URL doesn't count as big :P).

On mer side there are usually no dependencies from hybris-hal to libhybris, which complicates things. A split would help, and would fit better in my opinion since: "repositories which are required to be built with both Android.mk and Makefile.am are kind of weird ;)".

There is a workaround so if someone is absolutely against this, please let me know.

PR #406, opened 2019-01-22
Remove compat folder from libhybris

krnlyng:
As mentioned in #389 we would propose to split libhybris into two parts, the parts which are built into a package directly (linker, etc., i.e. stuff that's linked against glibc) and the parts which are needed to build compat libraries like libhwc2_compat_layer (Android.mk libraries, i.e. things linked against bionic) in order to simplify automated builds.

This commit drops the compat folder from libhybris and this would be a repository containing only the compat folder: https://github.com/krnlyng/libhybris-compat which we could move to https://github.com/libhybris/libhybris-compat if this change is ok for everyone.

Please let me know if this breaks a specific workflow so we can figure out a solution which works for everyone.


Tofee:
No problem from LuneOS's side, and separating the Android build and the "glibc" build makes sense.

bhush9:
LGTM from Halium side.

mariogrip:
LGTM from UBports side

FlorentRevest:
Alright for AsteroidOS as well

krnlyng:
Thanks! I will leave this open until tomorrow in case somebody notices a problem last minute :). The compat repo will be located here: https://github.com/libhybris/libhybris-compat

krnlyng:
Well there would be some header duplication which we also want to avoid so i will move this a bit behind.

PR #407, opened 2019-02-07
Android 8 support

krnlyng:
This works on devices < android o and adds support for android o.


Tofee:
A test as-is on my "rosy" (aarch64, Halium 7.1, Xiaomi Redmi 5) fails with:
`rosy luna-next[4608]: FATAL: 06:34:58.604: QPA-HWC: hw_get_module(HWC_HARDWARE_MODULE_ID, (const hw_module_t **)(&hwc_module)) in create returned -2`

I'll have to test a bit more to track where the issue comes from, but let's be careful about this PR for now.

hybris/common/hooks.c
```diff
@@ -2935,21 +3092,26 @@ static void* __hybris_get_hooked_symbol(const char *sym, const char *requester)
...
+    sdk_version = get_android_sdk_version();
+
+#if defined(WANT_LINKER_MM) || defined(WANT_LINKER_N) || defined(WANT_LINKER_O)
+    if (sdk_version > 21)
         found = bsearch(&key, hooks_mm, HOOKS_SIZE(hooks_mm), sizeof(hooks_mm[0]), hook_cmp);
 #endif
+    // make sure to skip the property hooks only when o.so is actually loaded
+    // since for testing and we sometimes set things like 99 as sdk version.
+    // The o linker is loaded when sdk_version >= 27 and exists.
+    if (!found && sdk_version < 24)
```
Tofee: Shouldn't it be < 27, here ? Otherwise it won't work with N linker.

krnlyng: this is related to https://github.com/libhybris/libhybris/pull/407/commits/bae4f72d935cdda72479b8d02285a8fdb6b8b98e

i need to wrap my head around this but basically it should work because of this: 
https://github.com/libhybris/libhybris/pull/407/commits/bae4f72d935cdda72479b8d02285a8fdb6b8b98e#diff-cf1210eb0ad9879f7d812c63baf11fb4R3386

krnlyng: @tofe can you check the output of "/usr/bin/getprop ro.build.version.sdk"

Tofee: It returns nothing in with "< 24", and it returns "25" with my little patching

krnlyng: i have reverted the patch since it breaks compatibility with halium

it's currently necessary on android 8 adaptations so i kept it for those, but we can look at that issue later

mkosola:
LGTM

krnlyng:
halium and asteroidos have given ok so i will merge this :)

PR #412, opened 2019-03-14
jb/mm/n: HYBRIS_ENABLE_LINKER_DEBUG_MAP enables Android libs in gdb

Tofee:
There are still some conflicts between android libs and glibc libs in the
gdb lib map.

Very likely, in most cases, the developer only wants to debug the glibc
related code, and is not interested in seeing the Android stack.
Therefore, introduce an env. variable and disable linking of android
libs by default.

### DIFF
--- a/hybris/common/jb/linker.c
+++ b/hybris/common/jb/linker.c
@@ -172,8 +172,28 @@ static struct link_map *r_debug_tail = 0;
 
 static pthread_mutex_t _r_debug_lock = PTHREAD_MUTEX_INITIALIZER;
 
+static int _is_android_debug_enabled() 
+{
+  static int _hybris_enable_android_debug = -1; // -1: not initialized
+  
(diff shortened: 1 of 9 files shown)


hybris/common/jb/linker.c
```diff
@@ -172,8 +172,28 @@ static struct link_map *r_debug_tail = 0;
 
 static pthread_mutex_t _r_debug_lock = PTHREAD_MUTEX_INITIALIZER;
 
+static int _is_android_debug_enabled()
```
krnlyng: lgtm but i think you could move the _is_android_debug_enabled function to hooks.c (where the linkers are loaded and pass the debug-enabled/disabled state via android_linker_init)
then we would avoid having the same code multiple times

Tofee: It should look better now!

PR #418, opened 2019-04-08
common: Ignore HYBRIS_LINKER_DIR for processes which should be treated securely.

krnlyng:


Tofee:
Looks fine to me; it has no LuneOS impact.

PR #419, opened 2019-06-01
Upstreaming changes from Halium

fredldotme:
This work includes:
- Update to the compat layer for Android >= 7.1
- Overriding EGL platform directory using environment variable `HYBRIS_EGLPLATFORM_DIR`
- Hooking `__set_errno_internal`


fredldotme:
Addressed in commit 0a94128537be7a6a451aec57f4f97839a70065db

Herrie82:
@krnlyng Are you OK to merge this now?

compat/media/camera_service.cpp
```diff
@@ -45,7 +45,7 @@ int main(int argc, char** argv)
     // for creating a new IMediaRecorder (MediaRecorder) instance over Binder
     MediaRecorderFactory::instantiate();
     // Enable audio recording for camera recording
-    CameraRecordService::instantiate();
+    // CameraRecordService::instantiate();
```
z3ntu: These changes aren't pretty

krnlyng: good point, why are these changes necessary?

fredldotme: I'll ask the author of the patch set but a quick test build without the line being commented resulted in failure due to `use of undeclared identifier 'CameraRecordService';`.

mlehtima: The header was removed in commit https://github.com/libhybris/libhybris/commit/ffcf03a248ad739f83ae55966da9c0599b0613d8 but I couldn't find the header or other files anywhere in Android source tree and web searches didn't help either, those files were only in libhybris.

fredldotme: How should we proceed here? Any specific info you need?

krnlyng: this probably needs to be inside some ifdefs?

z3ntu: halium-5.1 has `camera_record_service.cpp` in https://github.com/Halium/android_frameworks_av/tree/halium-5.1/media/libmedia but halium-7.1 doesn't have that. It rather sounds like these changes have to applied to the halium-7.1 as well instead of patching the libhybris side

Herrie82: camera_record_service.h Seems to be Ubuntu specific? All (5) references point to that? Seeing it's compat, isn't this Android 5.1 only anyway? https://github.com/search?q=CAMERA_RECORD_SERVICE_H&type=Code

mlehtima: Yes, based on copyright header in the files those were added by Canonical.

compat/media/camera_service.cpp
```diff
@@ -23,7 +23,7 @@
 #include "media_recorder_factory.h"
 #include "media_recorder.h"
 
-#include <media/camera_record_service.h>
+//#include <media/camera_record_service.h>
```
Herrie82: I would make this include conditional with #if ANDROID_VERSION_MAJOR==5

mlehtima: That would only work for Halium or other projects using Ubuntu based Android repos.

fredldotme: Is Halium/UBports the only consumer of the compat portions?

krnlyng: @fredl

> Is Halium/UBports the only consumer of the compat portions?

Some compat layers are also used by SailfishOS but this one isn't.

fredldotme: Should we remove it instead of commenting out since the header isn't in the tree anymore?

minlexx: > pmOS is using mainline kernels.

pmOS does not limit device ports to specific kernel, some ports are using downstream android kernels and Halium, too.

Herrie82: @minlexx In that case a build flag would make sense to me to allow for flexibility by the different OS-es using libhybris in their own way.

krnlyng: a build time flag would make sense to me, in theory you could even check for the existence of the header(s) in configure.ac like we did with HAS_HWCOMPOSER2_HEADERS and similar ones.

fredldotme: How does it look now? The condition can be fulfilled both via configure and from the Android side using `CAMERA_SERVICE_WANT_UBUNTU_HEADERS := 1`.

PR #423, opened 2019-08-28
extract-headers.sh: Use version from Mer (Adds Android 8 & 9 support)

Herrie82:
Mer's version from https://github.com/mer-hybris/droid-hal-device/blob/master/helpers/extract-headers.sh add support for Android 8 & 9 (Oreo & Pie) as well as some error handling. Since Halium and projects using Halium are looking into Android 8 & 9 based ports, it's good to update the script to cover these headers as well.

Signed-off-by: Herman van Hazendonk <github.com@herrie.org>

### DIFF
--- a/utils/extract-headers.sh
+++ b/utils/extract-headers.sh
@@ -1,61 +1,89 @@
-#!/bin/sh
+#! /bin/sh
 
-ANDROID_ROOT=$1
-HEADERPATH=$2
-MAJOR=$3
-MINOR=$4
-
-PATCH=$5
-PATCH2=$6
-PATCH3=$7
-
+error() {
+    echo "Error: $1" >&2
+}
 usage() {
-    echo "Usage: extract-headers.sh <ANDROID_ROOT> <HEADER_PATH> [Android Platform Version]"
+    echo "Usage: extract-headers.sh [Options] <ANDROID_ROOT> <HEADER_PATH>"
     echo
     echo "  ANDROID_ROOT: Directory containing the Android source tree."
     echo "  HEADER_PATH:  Where the headers will be extracted to."
     echo
-    echo "Android Platform Version:"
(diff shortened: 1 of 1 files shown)


utils/extract-headers.sh
```diff
@@ -1,61 +1,89 @@
...
-
-ANDROID_ROOT=$1
-HEADERPATH=$2
-MAJOR=$3
-MINOR=$4
-
-PATCH=$5
-PATCH2=$6
-PATCH3=$7
+#! /bin/sh
```
minlexx: A space before `/bin/sh` ? :open_mouth:

JL2210: Not to butt in here, but it does help when `grep`ping for `#!/bin/sh`.

Herrie82: I just copy & pasted from Mer. No code from me ;) Happy to change it either way.

Tofee: Well spaces in shebang is actually completely ok; see also https://unix.stackexchange.com/questions/276751/is-space-allowed-between-and-bin-bash-in-shebang for some more details.
As it would let us have an identical version as Mer's, it could be a good point.

krnlyng: should we fix this in the mer version then? also is there any better way how we can keep those two files in sync?

krnlyng:
has this been tested on halium side?

Herrie82:
@krnlyng There is no official Android 8/9 support for Halium yet. Seems to be OK for unofficial WIP one for Android 8.1. To answer the other question to keep things in sync. There seems to be some workarounds to allow files as sub-modules somehow. I'll have a look into that as well.

krnlyng:
@Herrie82 i think we shouldn't add a submodule to libhybris, but it's not nice to have this in two places, maybe Halium could use droid-hal-device as a submodule? Then we could remove the script from here. Otherwise i guess we can merge this.

PR #424, opened 2019-09-02
eglplatform_wayland: load libEGL at runtime

minlexx:
`hybris_egl_display_get_mapping()`  function is implemented
in libhybris's libEGL.so. However, eglplatform_wayland.so
is not linked to libEGL directly, causing undefined symbol
errors during loading, if libEGL was not already loaded by
some other dependencies. Therefore, we should try
to load libEGL at runtime here and resolve this function
dynamically.

Error is seen at startup of Plasma Mobile in postmarketOS, for example:
```
klte:~$ cat /tmp/plasmashell_logs 
Error: could not determine $DISPLAY.
org.kde.kwindowsystem: Loaded plugin "/usr/lib/qt5/plugins/kf5/org.kde.kwindowsystem.platforms/KF5WindowSystemKWaylandPlugin.so" for platform "wayland"
Using Wayland-EGL
ERROR: /usr/lib/libhybris/eglplatform_wayland.so
        Error relocating /usr/lib/libhybris/eglplatform_wayland.so: hybris_egl_display_get_mapping: symbol not found
Assertion failed: 0 (ws.c: _init_ws: 49)
KCrash: Attempting to start /usr/bin/plasmashell from kdeinit
Warning: connect() failed: : Connection refused
KCrash: Attempting to start /usr/bin/plasmashell directly
KCrash: Application 'plasmashell' crashing...
org.kde.kwindowsystem: Loaded plugin "/usr/lib/qt5/plugins/kf5/org.kde.kwindowsystem.platforms/KF5WindowSystemKWaylandPlugin.so" for platform "wayland"
Using Wayland-EGL
ERROR: /usr/lib/libhybris/eglplatform_wayland.so
        Error relocating /usr/lib/libhybris/eglplatform_wayland.so: hybris_egl_display_get_mapping: symbol not found
Assertion failed: 0 (ws.c: _init_ws: 49)
```

Some googling and talking to people reveals that this problem is known at least since year 2017:

```
Jan 23 17:01:21 <NotKit> Using Wayland-EGL … ERROR: /usr/lib/libhybris/eglplatform_wayland.so …
    Error relocating /usr/lib/libhybris/eglplatform_wayland.so: hybris_egl_display_get_mapping: symbol not found
Jan 23 17:01:25 <NotKit> oh, that famous issue
Jan 23 17:01:53 <bhushanshah> huh?
Jan 23 17:02:18 <NotKit> I think that also happened on Arch
Jan 23 17:02:34 <bhushanshah> do you know solution?
Jan 23 17:04:52 <NotKit> I remember that it required linker arguments change during libhybris build
Jan 23 17:05:24 <NotKit> used export LD_PRELOAD=/usr/lib/libEGL.so.1.0.0 as quick hack
```
http://www.merproject.org/logs/%23sailfishos-porters/%23sailfishos-porters.2017-10-01.log.html :
```
krnlyng    bshah, try adding -lEGL to LDFLAGS of hybris/egl/platforms/wayland/Makefile.am
```

But suggested solutions are not real solutions, but only a workarounds. True solution is using `dlopen()`, because as I understand, platform plugins are not supposed to be linked directly to egl implementation. And by the way, in libhybris buildsystem libEGL itself is built *after* egl_platform plugins.

-------------------------------------------------
Bug was originally reported at https://gitlab.com/postmarketOS/pmaports/issues/347

### DIFF
--- a/hybris/egl/platforms/wayland/eglplatform_wayland.cpp
+++ b/hybris/egl/platforms/wayland/eglplatform_wayland.cpp
@@ -36,6 +36,7 @@
 #include <unistd.h>
 #include <assert.h>
 #include <stdlib.h>
+#include <dlfcn.h>
 extern "C" {
 #include <eglplatformcommon.h>
 };
@@ -60,6 +61,13 @@ static __eglMustCastToProperFunctionPointerType (*_eglGetProcAddress)(const char
 static EGLSyncKHR (*_eglCreateSyncKHR)(EGLDisplay dpy, EGLenum type, const EGLint *attrib_list) = NULL;
 static EGLBoolean (*_eglDestroySyncKHR)(EGLDisplay dpy, EGLSyncKHR sync) = NULL;
 static EGLint (*_eglClientWaitSyncKHR)(EGLDisplay dpy, EGLSyncKHR sync, EGLint flags, EGLTimeKHR timeout) = NULL;
+
+/* The following function is implemented in libhybris's libEGL.so.
+ * However, eglplatform_wayland.so is not linking to libEGL directly,
+ * causing undefined symbol errors during loading, if libEGL was not
+ * already loaded by some other dependencies. Therefore, we should try
+ * to load libEGL at runtime here and resolve this function dynamically */
+typedef struct _EGLDisplay *(*PFNHYBRISEGLDISPLAYGETMAPPINGPROC)(EGLDisplay dpy);
 
 struct WaylandDisplay {
 	_EGLDisplay base;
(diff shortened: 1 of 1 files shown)


hybris/egl/platforms/wayland/eglplatform_wayland.cpp
```diff
@@ -211,7 +219,30 @@ extern "C" wl_buffer *waylandws_createWlBuffer(EGLDisplay dpy, EGLImageKHR image
 	    return NULL;
 	}
 	if (img->target == EGL_WAYLAND_BUFFER_WL) {
-		WaylandDisplay *wdpy = (WaylandDisplay *)hybris_egl_display_get_mapping(dpy);
+		/* load hybris's libEGL at runtime to call hybris_egl_display_get_mapping() */
```
krnlyng: i think it would be nicer if this happened in a separate function

minlexx: makes sense, separated wrapper code

minlexx:
So, I've moved dlopen wrapper into a separate function and tested that it actually works without LD_PRELOAD hack on my port [fresh vid](https://www.youtube.com/watch?v=1kXpkorNCik), [old vid](https://www.youtube.com/watch?v=v0iI0p6qKKw)

P.S. Mentioning other people that might be interested ( @bhush9 @NotKit )

hybris/egl/platforms/wayland/eglplatform_wayland.cpp
```diff
@@ -202,6 +210,36 @@ extern "C" int waylandws_post(EGLNativeWindowType win, void *buffer)
...
+ * Loads libhybris's libEGL at runtime to call hybris_egl_display_get_mapping()
+ */
+static struct _EGLDisplay *_hybris_egl_display_get_mapping(EGLDisplay dpy)
+{
+	static void *libEGL_handle = NULL;
+	static PFNHYBRISEGLDISPLAYGETMAPPINGPROC hybris_egl_display_get_mapping_fn = NULL;
+
+	if (!libEGL_handle) {
+		dlerror();  // cleanup error buffer
+		libEGL_handle = dlopen("libEGL.so.1", RTLD_NOW | RTLD_GLOBAL);
```
krnlyng: any reason why this isn't RTLD_LAZY?

minlexx: No specific reason, I just thought it would be safer. Is `RTLD_LAZY` better (faster)? I can try with `RTLD_LAZY` and if you think it's better I will fix.

PR #427, opened 2019-11-02
eglplatform: increase eglextensionsbuf

rinigus:
Increased to fit the extensions on Sony Tama/Akari.

On that device, I had maximal length of strings equal to

```
eglplatformcommon_eglQueryString max string length: 1060
waylandws_eglQueryString max string length: 1133
```

### DIFF
--- a/hybris/egl/platforms/common/eglplatformcommon.cpp
+++ b/hybris/egl/platforms/common/eglplatformcommon.cpp
@@ -335,8 +335,8 @@ extern "C" const char *eglplatformcommon_eglQueryString(EGLDisplay dpy, EGLint n
 	if (name == EGL_EXTENSIONS)
 	{
 		const char *ret = (*real_eglQueryString)(dpy, name);
-		static char eglextensionsbuf[1024];
-		snprintf(eglextensionsbuf, 1022, "%sEGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer %s", ret ? ret : "",
+		static char eglextensionsbuf[2048];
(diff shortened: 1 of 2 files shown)


krnlyng:
should we dynamically allocate the buffer instead the first time this function is called? and unallocate once the process exits? (atexit)

rinigus:
@krnlyng : I cannot comment on that, as I don't even know where it is called from specifically nor how frequently its done. As far as I could see, increase in the buffer size made it possible to run on XZ2. Hopefully, someone knowledgeable in the area will step in.

krnlyng:
> @krnlyng : I cannot comment on that, as I don't even know where it is called from specifically nor how frequently its done. As far as I could see, increase in the buffer size made it possible to run on XZ2. Hopefully, someone knowledgeable in the area will step in.

sorry i meant within the function not from where it is called

rinigus:
If we want to deallocate it on process exit, we will need to signal it somehow from outside. If I understand you correctly. Alternative would be to have a static pointer and allocate the buffer which is sufficiently large to fill the string. Later, as running proceeds, reallocate to make it larger if needed. 

Not sure whether this is a great solution as it does add complexity to the whole process and, I suspect, will make it slower on average due to multiple allocations.

krnlyng:
We could do it like in mesa:
https://github.com/mesa3d/mesa/blob/fa611b07f1115197916b12bd69200315f6df21af/src/egl/main/eglglobals.c#L128
https://github.com/mesa3d/mesa/blob/fa611b07f1115197916b12bd69200315f6df21af/src/egl/main/eglapi.c#L690

The value shouldn't change after the first call to eglQueryString so we can allocate it once inside the function with malloc and otherwise return the value we have stored (in a global variable).

To deallocate the buffer we can use atexit:
http://man7.org/linux/man-pages/man3/atexit.3.html

to run a function on exit of the program which free's the buffer.

PR #429, opened 2019-11-14
hybris: common: o: improve android pie support

krnlyng:
We don't have an actual linker for android 9 based ports, the android 8 linker works well enough, however an extra include path is needed for some libraries in /system/lib{64}/vndk-28.


JamiKettunen:
Works nice for me using Sailfish OS based on `hybris-16.0` (Pie) :+1: no more need for my hacky symlink to get working audio https://github.com/sailfishos-oneplus5/droid-config-cheeseburger/commit/c371cf17bd0d961597d8411819577c609bca0e7d
Any chance for this to get merged soon? I've been cherry-picking the commit locally for a bit :eyes:

PR #430, opened 2019-12-06
hybris: egl: Update EGL headers to version 1.5.

mlehtima:


mlehtima:
I removed the implementation part of the code because according to Android documentation only Android 10 has support for EGL 1.5 and therefore I have no way to test that yet. Actual implementation of EGL 1.5 can be added later.

PR #435, opened 2020-01-19
Fallback to own property implementation if libc is not present

mariogrip:
This adds fallback to own property implementation if libc is not present
instead of aborting. This is needed as libhybris may be linked on non
hybris devices where we use getprop to determine what device type we are
running on.

### DIFF
--- a/hybris/properties/hybris_properties.c
+++ b/hybris/properties/hybris_properties.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018 Jolla Ltd. <franz.haider@jolla.com>
+ * Copyright (c) 2020 UBports foundation <marius@ubports.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


hybris/properties/hybris_properties.c
```diff
@@ -61,20 +68,29 @@ int property_list(void (*propfn)(const char *key, const char *value, void *cooki
...
+        return my_property_get(key, value, default_value);
 }
 
 int property_set(const char *key, const char *value)
 {
     ensure_bionic_properties_initialized();
 
-    return bionic_property_set(key, value);
+    if (!own_impl)
+         return bionic_property_set(key, value);
```
krnlyng: Indentation seems to be off here (tabs vs spaces?)

krnlyng: sorry for the nitpick, can you fix the indentation here? then i'll merge it

mariogrip: Oh sorry, this got buried in my email. Will fix it right away.

PR #440, opened 2020-03-15
egl: wayland: Remove additional space before EGL_EXT

JBBgameich:
The string returned by eglplatformcommon_eglQueryString already ends with a space,
adding an additional one causes crashes (tested on a PowerVR device).

Before:
An additional space was visible before EGL_EXT_swap_buffers_with_damage.

After:

This probably needs verification on other devices

### DIFF
--- a/hybris/egl/platforms/common/eglplatformcommon.cpp
+++ b/hybris/egl/platforms/common/eglplatformcommon.cpp
@@ -336,13 +336,7 @@ extern "C" const char *eglplatformcommon_eglQueryString(EGLDisplay dpy, EGLint n
 	{
 		const char *ret = (*real_eglQueryString)(dpy, name);
 		static char eglextensionsbuf[2048];
-		snprintf(eglextensionsbuf, 2046, "%sEGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer %s", ret ? ret : "",
-#ifdef WANT_WAYLAND
-			"EGL_WL_bind_wayland_display "
-#else
-			""
-#endif
-		);
+		snprintf(eglextensionsbuf, 2046, "%sEGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer EGL_WL_bind_wayland_display", ret ? ret : "");
 		ret = eglextensionsbuf;
 		return ret;
 	}


fcipaq:
This might also be a bug in the PVR driver ending the returned string with a trailing whitespace. KHR documents states the extensions are separated with spaces, but does not state that there has to be a trailing whitespace.
Maybe it's best to check the returned string for a trailing whitespace and if it exists cut if off.

fcipaq:
Just learned, this is not PVR related. Trailing whitespace comes from here:
https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/common/eglplatformcommon.cpp#L341

krnlyng:
Can you describe a bit more where the crash happens? i haven't observed any of that nature so far and would be interested to know.

eglplatformcommon_eglQueryString seems to always return a space at the end and i don't see a reason for that (but of course that needs to be checked), maybe we should fix eglplatformcommon_eglQueryString instead?

fcipaq:
KHR documentation states: "EGL_EXTENSIONS Returns a space separated list of supported extensions to EGL. If there are no extensions, then the empty string is returned." So no trailing whitespace.

I would suggest to extend the list by "%s EXT_extension" rather than "%sEXT_extension ".

fcipaq:
For stack trace of a crash please see the screenshot (thanks to JBBgameich)

photo_2020-03-18_08-25-43

I am unsure, why some programs are affected whereas others are not. Maybe it is an additional bug in those programs. One theory might be that "EXT_1  EXT_2" (two whitespaces) might be interpreted by the the calling program as "EXT_1", " EXT_2" OR "EXT_1", "", "EXT_2". So this is undefined behavior.

fcipaq:
My suggestion would be to check if "ret" has a trailing whitespace here:

https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/common/eglplatformcommon.cpp#L337
 
if that is the case, concatenate "EGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer" otherwise concatenate "EGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer"

And from there always concatenate " EGL_EXT_xyz" here:
https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/common/eglplatformcommon.cpp#L341

and here:

https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/wayland/eglplatform_wayland.cpp#L290

fcipaq:
And maybe (while we're at it) fix the dublicate "#ifdef WAND_WAYLAND" here:
https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/common/eglplatformcommon.cpp#L334

and here:

https://github.com/libhybris/libhybris/blob/master/hybris/egl/platforms/common/eglplatformcommon.cpp#L340

JBBgameich:
Will look into reworking the merge request tomorrow, thanks for your comments / help so far!

krnlyng:
I think the innner ifdef should have been removed, not the outer, since now EGL_HYBRIS_WL_acquire_native_buffer will always be included.

krnlyng:
i would suggest using:
`		snprintf(eglextensionsbuf, 2046, "%sEGL_HYBRIS_native_buffer2 EGL_HYBRIS_WL_acquire_native_buffer EGL_WL_bind_wayland_display", ret ? ret : "");`
since that is effectively the same

otherwise i think we can merge this after squashing the commits, thanks!

PR #441, opened 2020-03-21
hybris: hwc2: avoid crash in hwc2_compat_display_present

NotKit:
Check if presentFence is a valid pointer before calling dup(), which can happen if HWC2::Display::present() fails for some reason.

Tested to fix random lipstick crashes on Redmi Note 7 (lavender).

### DIFF
--- a/compat/hwc2/hwc2_compatibility_layer.cpp
+++ b/compat/hwc2/hwc2_compatibility_layer.cpp
@@ -214,7 +214,11 @@ hwc2_error_t hwc2_compat_display_present(hwc2_compat_display_t* display,
     android::sp<android::Fence> presentFence;
     HWC2::Error error = display->self->present(&presentFence);
 
-    *outPresentFence = presentFence->dup();
+    if (presentFence != NULL) {
+        *outPresentFence = presentFence->dup();
+    } else {
+        *outPresentFence = -1;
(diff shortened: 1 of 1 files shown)


compat/hwc2/hwc2_compatibility_layer.cpp
```diff
@@ -214,7 +214,11 @@ hwc2_error_t hwc2_compat_display_present(hwc2_compat_display_t* display,
     android::sp<android::Fence> presentFence;
     HWC2::Error error = display->self->present(&presentFence);
 
-    *outPresentFence = presentFence->dup();
+    if (presentFence) {
```
0312birdzhang: On my build environment, i need change to `if (presentFence != NULL) {`.
Tested on vince, no lipstick crashes.

krnlyng: > On my build environment, i need change to `if (presentFence != NULL) {`.
> Tested on vince, no lipstick crashes.

Does that also work for you @NotKit? if so maybe we can change it so it builds everywhere and merge this.

NotKit: Tested to work for me. I modified the commit, let's merge this then.

PR #361, opened 2017-11-01
[WIP] introduce glesv3 wrapper

bhush9:
RFC, DONT MERGE YET.

This uses the scripts in utils to generate the wrapper.. Not quite sure how to test this yet.

PR #445, opened 2020-04-24
Add OpenGL ES 3 support

mlehtima:
I also added common test functions for the most commonly shared shader compilation and hwcomposer code. After the graphics test fixes I removed test_hwc2 because now test_hwcomposerworks for all hwcomposer versions. Supersedes https://github.com/libhybris/libhybris/pull/361. Compared to the previous OpenGL Es 3.0 pull request this also supports versions 3.1 and 3.2 and uses a simpler approach for implementing the glesv3 support inside glesv2 code.


hybris/egl/egl.c
```diff
@@ -425,14 +425,12 @@ __eglMustCastToProperFunctionPointerType eglGetProcAddress(const char *procname)
 			ret = _hybris_libgles1 ? dlsym(_hybris_libgles1, procname) : NULL;
 			break;
 		case 2:  // OpenGL ES 2.0 API
+		case 3:  // OpenGL ES 3.x API
```
krnlyng, line 428: i know that on android libGLESv3 is usually the same as libGLESv2 but should we still load libGLESv3 in the glesv3 case just in case that changes?

hybris/tests/test_glesv2.cpp
```diff
@@ -94,14 +86,24 @@ int main(int argc, char **argv)
...
 
-	display = eglGetDisplay(NULL);
+	HWComposer *win = create_hwcomposer_window();
+
+	if (!win) {
+		printf("Failed to create native window\n");
+		return 1;
+	}
+	printf("created native window\n");
+	/* Not needed with hybris window platform
```
krnlyng, line 103: is the comment still necessary?

hybris/tests/test_glesv3.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+	HWComposer *win = create_hwcomposer_window();
+
+	if (!win) {
+		printf("Failed to create native window\n");
+		return 1;
+	}
+	printf("created native window\n");
+	/* Not needed with hybris window platform
+	 * hybris_gralloc_initialize(0); */
```
krnlyng, line 105: same as above?

hybris/tests/test_glesv2.cpp
mlehtima: Probably not needed, I can remove that.

Issue #443, opened 2020-04-06
Race condition when using SDL2/Wayland


fcipaq:
I am using libhybris on a x86 Android tablet with PowerVR graphics to run KWin-Wayland with Plasma desktop (Debian). The desktop is working well.
I am trying to use a program, namely "Neverball", a (famous?) opensource game from the Debian repo, with SDL2/wayland support.

Program start fails with 

`gralloc.c:132: hybris_gralloc_initialize: Assertion `NULL' failed.`

Now, this is due to two calls the SDL lib makes to libhybris:

1. `eglQueryString` which then calls `_init_ws();` in `ws.c`
2. While still in `_init_ws();` libhybis gets a second call to `GetDisplay` and again, `_init_ws();` is being calld.
3. Because the first call to `_init_ws();` has not yet completed causes a double call on `hybris_gralloc_initialize` eventually leading to the error message above.

I am not sure if this a SDL2 bug or libhybris bug, but a solution that worked for me, was a mutex approach in `ws.c`:

```
static struct ws_module *ws = NULL;
static int ws_init_locked = 0;

static void _init_ws()
{
	do {} while (ws_init_locked)

	if (ws == NULL)
	{
         ws_init_locked = 1;
         ...
         ws_init_locked = 0;
	}
}

```
In fact I am unsure why there even is a call from the SDL2 side to `GetDisplay` before `eglQueryString` has finished...

fcipaq:
Another solution that worked for me is to make multiple calls to `hybris_gralloc_initialize` non fatal. I.e. replace the assertion by a warning:

https://github.com/libhybris/libhybris/blob/master/hybris/gralloc/gralloc.c#L132

I do not see any harm here as the code does in case `version != -1`

krnlyng:
looks like a libhybris bug: "EGL and its client APIs must be threadsafe." from https://www.khronos.org/registry/EGL/specs/eglspec.1.4.withchanges.pdf

fcipaq:
Thanks, that sound reasonable. I'll test it and if it works, I'll MR.

krnlyng:
@fcipaq i guess we can close this now?

PR #448, opened 2020-04-27
Prevent race condition on multi thread init of ws. Solves #443

fcipaq:
These commits solve two issues:

1. Solves #443 where a multi threaded call to `_init_ws` (while another thread has not yet completed `_init_ws`) leads to a second call of `hybris_gralloc_initialize` which then fails because it has already completed the initialization of the gralloc module.

2. Solves subsequent crashes of `kdeinit` because `HYBRIS_EGLPLATFORM` is reset to "" but it is not unset (it's still defined). This leads to `_init_ws` trying to load the file `eglplatform_.so` which obviously does not exist.
Therefore, fallback to the default EGL platform in case the environment variable is defined but empty.

### DIFF
--- a/hybris/egl/ws.c
+++ b/hybris/egl/ws.c
@@ -22,6 +22,7 @@
 #include <stdio.h>
 #include <sys/auxv.h>
 #include <pthread.h>
+#include <string.h>
 
 pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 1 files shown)


hybris/egl/ws.c
```diff
@@ -21,13 +21,23 @@
...
+pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 
 static struct ws_module *ws = NULL;
 
 static void _init_ws()
 {
 	if (ws == NULL)
 	{
+		pthread_mutex_lock(&mutex);
+		if (ws != NULL) {
```
krnlyng, line 36: how about moving the pthread_mutex_(un)lock in the scope of _init_ws i.e. move it out from if (ws == NULL) to avoid that extra if (ws != NULL) check?

fcipaq:
The reason I did not do this is because this will block `_init_ws` even when the initialization has completed. I mean it causes a block-unblock cycle on any call to the window system. If this does not have any impact on performance, I guess it's ok then to call it at the beginnnig and the end.

PR #449, opened 2020-04-29
Code cleanup

mlehtima:
Silence a lot of compiler warnings. Move to using wayland-egl from wayland project but keep our version of that in case some projects use older wayland.


hybris/common/hooks.c
```diff
@@ -257,6 +257,7 @@ static int __android_pthread_cond_pulse(android_cond_t *cond, int counter)
     fret = syscall(SYS_futex , &cond->value,
                    pshared ? FUTEX_WAKE : FUTEX_WAKE_PRIVATE, counter,
                    NULL, NULL, NULL);
+    (void)fret;
     LOGD("futex based pthread_cond_*, value %d, counter %d, ret %d",
                                             cond->value, counter, fret);
     return 0;
```
krnlyng, line 263: not related to this PR but here we return always 0, is that intended?

hybris/common/hooks.c
```diff
@@ -1591,8 +1592,9 @@ static int _hybris_hook_fsetpos(FILE *fp, const bionic_fpos_t *pos)
 
     fpos_t my_fpos;
     my_fpos.__pos = *pos;
-    memset(&my_fpos.__state, 0, sizeof(mbstate_t));
-    mbsinit(&my_fpos.__state);
+    if (!mbsinit(&my_fpos.__state)) {
```
krnlyng, line 1595: I think this should be just memset(&my_fpos.__state, 0, sizeof(mbstate_t)); without the entire mbsinit function which is used to check whether the variable is initialized but in this state we know it's uninitialized since we just created it.

PR #456, opened 2020-07-13
hybris: hook __cxa_thread_atexit (no wrap dtor)

peat-psuwit:
This function is used by the compiler to destruct `thread_local`
variables at thread's exit. Virtually no one uses it, but I planed to
use to solve libmedia_compat_layer's problem when using remote codec.

This implementation don't wrap dtor, so it cannot do any reference
counting decrease. However, I guess dlclosing is rare enough that we
can sustain a minor memory lose.

For an implementation that wraps dtor, see #457.


peat-psuwit:
Per discussion in #457, wrapping dtor is a better idea.

PR #457, opened 2020-07-13
hybris: hook __cxa_thread_atexit (wrap dtor)

peat-psuwit:
This function is used by the compiler to destruct `thread_local`
variables at thread's exit. Virtually no one uses it, but I planed to
use to solve libmedia_compat_layer's problem when using remote codec.

This implementation wraps dtor so that we can do reference counting.
This prevents memory leak when a library is dlclose'd, at the expense
of more memory usage and more code to execute.

For an implementation that doesn't wrap dtor, see #456.

### DIFF
--- a/hybris/common/hooks.c
+++ b/hybris/common/hooks.c
@@ -85,6 +85,8 @@ extern int my_property_list(void (*propfn)(const char *key, const char *value, v
 // this is also used in bionic:
 #define bool int
 
+#include "dso_handle_counters.h"
+
 #ifdef WANT_ARM_TRACING
 #include "wrappers.h"
 #endif
@@ -2244,16 +2246,15 @@ static void *_hybris_hook___get_tls_hooks()
 struct __wrapped_atexit {
     void (*dtor)(void *);
     void *obj;
-    void *dlopen_handle;
+    void *dso_handle;
 };
 
 static void __dtor_wrapper(void *obj) {
     struct __wrapped_atexit *wrapped = obj;
     /* Call the wrapped dtor. */
     wrapped->dtor(wrapped->obj);
-    /* Call dlclose, effectively decrease refcount. */
-    if (wrapped->dlopen_handle)
-        _android_dlclose(wrapped->dlopen_handle);
+    /* Reduce dso_handle's ref count. */
+    __hybris_remove_thread_local_dtor(wrapped->dso_handle);
     /* Free the wrapper. */
     free(wrapped);
 }
@@ -2263,35 +2264,21 @@ extern const void * const __dso_handle;
 static int _hybris_hook___cxa_thread_atexit(void (*dtor)(void *), void *obj,
                                             void *dso_symbol)
 {
-    /*
-     * Use android_dladdr() to find the library then use android_dlopen()
-     * to increase refcount. We then wrap the dtor with our own function
-     * which will call the dtor then call android_dlclose().
-     */
-    static __thread void *dso_symbol_cache;
-    static __thread Dl_info dso_symbol_info;
-    if (dso_symbol_cache != dso_symbol) {
-        dso_symbol_cache = dso_symbol;
-
-        if (!_android_dladdr(dso_symbol, &dso_symbol_info)) {
-            dso_symbol_info.dli_fname = NULL;
-        }
-    }
-
     struct __wrapped_atexit *wrapped = malloc(sizeof(struct __wrapped_atexit));
     wrapped->dtor = dtor;
     wrapped->obj = obj;
-    if (dso_symbol_info.dli_fname)
-        wrapped->dlopen_handle = _android_dlopen(dso_symbol_info.dli_fname,
-                                                 /* flags */ 0);
-    else
-        wrapped->dlopen_handle = NULL;
+    wrapped->dso_handle = dso_symbol;
 
     /* Call Glibc's implementation. Pass our symbol to prevent ourself from
      * being unloaded. */
     int ret;
-    if ((ret = __cxa_thread_atexit(__dtor_wrapper, wrapped, &__dso_handle)) != 0)
+    if ((ret = __cxa_thread_atexit(__dtor_wrapper, wrapped, &__dso_handle)) != 0) {
         free(wrapped);
+        return ret;
+    }
+
+    /* Increase refcount of this dso_symbol. */
+    __hybris_add_thread_local_dtor(dso_symbol);
 
     return ret;
 }
--- a/hybris/common/hooks.c
+++ b/hybris/common/hooks.c
@@ -2163,6 +2163,8 @@ int _hybris_hook_clearenv(void)
 
 extern int __cxa_atexit(void (*)(void*), void*, void*);
 extern void __cxa_finalize(void * d);
+extern int __cxa_thread_atexit(void (*dtor)(void *), void *obj,
+                               void *dso_symbol);
 
 struct open_redirect {
     const char *from;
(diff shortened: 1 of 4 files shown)


krnlyng:
Is this and #456 still a draft or ready? And do you have some information what/how we should test before merging this?

peat-psuwit:
The idea is, I'm not sure which approach is better between not wrapping destructor (#456) and wrapping it to make sure the library is unloaded (this PR). Thus, I've created both PR so that we can discuss pros & cons of both approaches.

The duty of this function is to ensure that the variable's destructor will be called when the thread is destructed and to make sure that the shared library containing the destructor is not unloaded before the destructor is called. That's why, in both PRs, the hook calls `dladdr()` and `dlopen()` to make sure the shared library stays loaded. However, in #456, I decided not to wrap the destructor, which leaves the library loaded even after all threads which use `thread_local` has exited. In this PR, I wrap the destructor to allow us to `dlclose()` too, which, given the library has been `dlclose()`'ed elsewhere, allow the library to be unloaded after all threads using that `thread_local` has exited.

You can learn more about this function at glibc's wiki [1]. For some prior art, Glibc keeps track of the library's call to this function internally, while bionic's implementation does nothing that prevents library unload.

For testing, this commit on my branch [2] is the code that makes this required. However, I guess libmedia_compat_layer is used by essentially no one but Ubuntu Touch. So, you probably need to add the following code to e.g. libdroidmedia, then `dlsym()` and call it from glibc code:

```cpp
void *test_thread_local() {
    static thread_local sp<MediaCodecInfo> codec_info = NULL;

    return codec_info;
}
```

Without either of the PR, it will crash inside bionic's implementation of the function.

[1] https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables
[2] https://github.com/peat-psuwit/libhybris/commit/3972a27ecc352eb4df81d374e3704da7872f037c

krnlyng:
```In this PR, I wrap the destructor to allow us to dlclose() too, which, given the library has been dlclose()'ed elsewhere, allow the library to be unloaded after all threads using that thread_local has exited.```

This sounds reasonable?

**Comment by peat-psuwit** [comment]
Pros/cons of #456:
+ A bit less memory usage for most of the use cases where the library isn't `dlclose()`'d.
+ Less complexity.
- The library will always stay loaded until the program terminates.
  - But, again, most of the time the library won't get `dlclose()`'d.
* Note that even most of the time the library isn't `dlclose()`'d, if it is we must not crash, which is the point of having either PR.

Pros/cons of #457 (this PR):
+ The library will eventually be unloaded when all threads using such `thread_local` terminates.
  - However, if the main thread accesses the `thread_local`, the library will most likely stay loaded until the program terminates anyway.
- More complexity.
- A bit of memory will be used for bookkeeping.

**Comment by krnlyng** [comment]
I would still argue that #457 sounds more reasonable even though it causes a little bit more memory usage for bookkeeping. I'm thinking that if the library is dlclosed for the last time it really should be unloaded so that destructors are all called properly etc.? Or is my understanding incorrect?

**Comment by peat-psuwit** [comment]
Hmm, you have a point about the library's destructor. If the main thread accesses the `thread_local` variable and then eventually calls `dlclose()`:
- #456 will leave the library loaded, as if there's another `dlopen()` call (because literally there is).
- #457 will eventually unload the library when the main thread exits, at which point the library's destructor will be called.

The question is: if a library is still loaded at the time the program/the main thread exit, will the library's destructor be eventually called somehow (in glibc/bionic/libhybris case)? If the answer is no, then I guess #457 actually makes more sense.

**Comment by krnlyng** [comment]
@peat-psuwit destructors will be called from here:

https://github.com/libhybris/libhybris/blob/30a137c54d4bc8f39612b55f66dee12209ca80b6/hybris/common/o/linker.cpp#L1821

which is being called from here:
https://github.com/libhybris/libhybris/blob/30a137c54d4bc8f39612b55f66dee12209ca80b6/hybris/common/o/linker.cpp#L2235

**Comment by krnlyng** [comment]
So i guess the answer is no, since soinfo_unload only calls the destructors if the last reference to the library is gone.

**Comment by peat-psuwit** [comment]
I've tested the (current) behavior for glib & bionic (Android 7 & Android 9) using [the test in this gist](https://gist.github.com/peat-psuwit/2ff18dfa2b980d8c3e8feab9ec9ee2dd). The shared library contains global `AAA`, `thread_local` `BBB` and lib's destructor. The table shows which destructors are called and in which order.

Run `dlclose()` / Has `thread_local`    | glibc | Android 7 | Android 9
------------------------------------------------- | ------ | -------- | -----------------------
Linked in / No `thread_local`               | at exit: lib -> AAA              | at exit: AAA | at exit: AAA
No `dlclose()` / No `thread_local`        | at exit: AAA -> lib             | at exit: AAA | at exit: AAA
Run `dlclose()` / No `thread_local`      | at `dlclose()`: lib -> AAA   | at `dlclose()`: lib -> AAA | at `dlclose()`: lib -> AAA
No `dlclose()` / Has `thread_local`      | at exit: BBB -> AAA -> lib | at exit: BBB -> AAA | at exit: BBB -> AAA
Run `dlclose()` / Has `thread_local`    | at exit: BBB -> AAA -> lib | at `dlclose()`: lib -> AAA. Crash at exit | at exit: BBB -> lib -> AAA

So, it seems like bionic linker/libc doesn't call the lib's destructor if it's not closed with `dlclose()`, which is on the contrary with glibc (notice that AAA's destructor is called in `__cxa_finalize()`, which is why they're called even if lib's destructor isn't). However, I just noticed that since Android 9, the `thread_local` crashing on exit has been fixed, which is good, but libhybris doesn't have Android 9 linker yet.

Currently, libhybris has the same behavior as bionic as it doesn't do anything special. This PR (#457) essentially implements the behavior of Android 9 but does so using public APIs of the linker, while the other PR (#456) act as of `dlclose()` hasn't been called.

So, IMO, this PR makes more sense, but I wonder if we should match Android 7's behavior on Android 7 and less though?

Side note:
- When we have Android 9+'s linker in, we can implement this PR in terms of the linker's private API too (and a quick glance suggest that it's faster).
- Indeed, after I've read Android's code, I think I can implement this even better. Only if I can use C++'s unordered map in hooks.c...

**Comment by peat-psuwit** [comment]
Turns out, by using a separate C++ file, I can use `unordered_map`!

File: hybris/common/hooks.c [hunk 1]
```diff
@@ -2239,6 +2243,46 @@ static void *_hybris_hook___get_tls_hooks()
...
+
+static int _hybris_hook___cxa_thread_atexit(void (*dtor)(void *), void *obj,
+                                            void *dso_symbol)
+{
+    struct __wrapped_atexit *wrapped = malloc(sizeof(struct __wrapped_atexit));
+    wrapped->dtor = dtor;
+    wrapped->obj = obj;
+    wrapped->dso_handle = dso_symbol;
+
+    /* Call Glibc's implementation. Pass our symbol to prevent ourselve from
```
mlehtima, line 2272: Nitpick: typo in ourselve, probably should be ourself.

krnlyng: @peat-psuwit can you fix the typo to satisfy @mlehtima , then i think we can merge it

peat-psuwit: Sorry, forgot about that. Will fix it soon.

peat-psuwit: Fixed.

mlehtima (approved):

PR #460, opened 2020-08-24
hooks: remove __attribute__((packed)) from struct bionic_file

NotKit:
Bionic does not have this attribute for struct __sFILE definition. Fixes struct size mismatch on aarch64 (used to be 144 vs 152 on Android), which resulted in crash of code using stdout/stderr symbols like fprintf(stderr, ...) found in MTK audio HAL.

Thanks to @krnlyng for noticing this.


krnlyng (approved):

mlehtima (approved):

mariogrip:
Oh! This have might the magic bug we have been fighting for a while on arm64! Will try this out! :D

PR #470, opened 2020-11-19
hooks: hook __cxa_atexit ourself to prevent segfault at exit

peat-psuwit:
If an Android library which place a __cxa_atexit() hook is not
dlclose()'ed before libhybris-common is unloaded, the hook will be
called when the program exit. This can cause segfault if the hook
refers to libhybris'es hooked symbol.

Record libraries that placed hooks, and then call them (through
__cxa_finallize) when we're exiting.

As hooks.c is growing, and because this is easier to implement in C++,
I've decided create a subdirectory specifically to contains hooks in the
future. Files in this subdirectory will be compiled with -fvisibility=
hidden to reduce global symbol pollution.

### DIFF
--- a/hybris/common/Makefile.am
+++ b/hybris/common/Makefile.am
@@ -1,5 +1,7 @@
 lib_LTLIBRARIES = \
 	libhybris-common.la
+noinst_LTLIBRARIES = \
+	hooks/libhybris-hooks.la
 
 SUBDIRS =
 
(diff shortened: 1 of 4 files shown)


krnlyng:
Hi, this looks quite nice, what is left to remove the draft status from this?

peat-psuwit:
The reason I leave this as a draft is that I thought there might be a better solution than this. Preferably I would like to dlclose() all libraries at exit, but I can't find a way to retrieve the list of opened libraries. If you agree that this approach is Ok, then I guess I can un-draft it.

PR #483, opened 2021-06-07
compat/media: fix handling of metadata in Android > 8

mardy:
The metadata buffer is held as a member variable by the MediaBuffer
class, which returns it as a reference. We should not make a copy of it,
or any change we make to the metadata will be affecting only our local
copy.
This commit ensures that we only keep a pointer to the metadata object
which is associated with the given buffer. We also need to increase the
reference count of the buffer for the lifetime of our class, to ensure
it does not get destroyed while we are using it.

Change-Id: I87c82decb099936d9d2ac79148c1f2a7b0e7a382

### DIFF
--- a/compat/media/media_meta_data_priv.h
+++ b/compat/media/media_meta_data_priv.h
@@ -26,11 +26,58 @@ struct MediaMetaDataPrivate
 public:
     static MediaMetaDataPrivate* toPrivate(MediaMetaDataWrapper *md);
 
+#if ANDROID_VERSION_MAJOR>=8
+    struct MetaDataPtr {
+        MetaDataPtr(android::MediaBufferBase *buf = nullptr):
+            buffer(buf)
+        {
+            if (buffer) {
+                buffer->add_ref();
+                data = &buffer->meta_data();
+            } else {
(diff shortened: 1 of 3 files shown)


compat/media/media_meta_data_priv.h
```diff
@@ -26,11 +26,58 @@ struct MediaMetaDataPrivate
...
+        android::MediaBufferBase *buffer;
+        android::MetaDataBase *data;
+    };
+#endif
+
     MediaMetaDataPrivate();
+#if ANDROID_VERSION_MAJOR>=8
+    MediaMetaDataPrivate(android::MediaBufferBase *buffer);
+#endif
     MediaMetaDataPrivate(const android::sp<android::MetaData> &md);
```
mlehtima, line 73: Should this (and the matching implementation) be conditionally enabled only for ANDROID_VERSION_MAJOR < 8 i.e. within else of the previous if? Based on my understanding of the code this is not used when ANDROID_VERSION_MAJOR>=8.

fredldotme:
I guess this can be closed by now.

PR #484, opened 2021-07-11
egl: rename eglDestroyImageKHR to _my_eglDestroyImageKHR to make apitrace work properly.

krnlyng:


krnlyng:
Without this apitrace gets confused and cannot override the eglDestroyImageKHR function.

mlehtima (approved):

PR #489, opened 2021-09-20
hybris: hooks: also pass symbols aquired though android_dlsym through get_hooked_symbol.

krnlyng:
This is to ensure that if a library dlsym's a libc symbol that we hook
doesn't end up calling the bionic libc variant instead of our hook.

Fixes nvidia targets.

Unfortunately at this point we do not know if it's a function or an object so we cannot pass it through create_wrapper for now.


mlehtima (approved):
Small typos in commit message and PR title (aquired -> acquired, though -> through). Otherwise looks good.

PR #490, opened 2021-09-26
hybris: hwc2: implement Device::destroyDisplay wrapper

g7:
Device::destroyDisplay was not exposed in the compat layer, thus clients had no way to destroy displays inside mDisplays, potentially leaving disconnected displays behind.


compat/hwc2/hwc2_compatibility_layer.cpp
```diff
@@ -132,6 +132,12 @@ hwc2_compat_display_t* hwc2_compat_device_get_display_by_id(
     return display;
 }
 
+void hwc2_compat_device_destroy_display(hwc2_compat_device_t* device,
```
NotKit, line 135: Would it make more sense to have hwc2_compat_device_destroy_display(hwc2_compat_device_t* device, hwc2_compat_display_t* display) so it also frees the compat struct and is in line with how hwc2_compat_display_destroy_layer works?

g7: It does indeed, and should be addressed. Thanks!

NotKit (approved):

PR #497, opened 2021-11-16
hybris: eglplatform_hwcomposer: allow creating multiple native windows

g7:
This pull request removes the arbitrary limit of one native window inside the hwcomposer platform.

### DIFF
--- a/hybris/egl/platforms/hwcomposer/eglplatform_hwcomposer.cpp
+++ b/hybris/egl/platforms/hwcomposer/eglplatform_hwcomposer.cpp
@@ -9,6 +9,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <assert.h>
+#include <mutex>
+#include <algorithm>
 extern "C" {
 #include <eglplatformcommon.h>
 };
@@ -17,7 +19,8 @@ extern "C" {
 
 #include <hybris/gralloc/gralloc.h>
 
-static HWComposerNativeWindow *_nativewindow = NULL;
+static std::vector<HWComposerNativeWindow *> _nativewindows;
+static std::mutex _nativewindows_mutex;
 
 extern "C" void hwcomposerws_init_module(struct ws_egl_interface *egl_iface)
 {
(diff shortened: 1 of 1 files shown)


krnlyng:
Hi, i don't quite understand why this is needed, can you elaborate?

g7:
Hello,

it's mostly useful on devices supporting multiple outputs, so that for example it's possible to create another native window for usage of an external HDMI display.

krnlyng:
I'm not sure if creating a seconday window is the right thing to do in that case, shouldn't you be doing something like this? https://github.com/mer-hybris/qt5-qpa-hwcomposer-plugin/pull/93

g7:
Hmm, thanks for the link. I have yet to do a through look at it (and on the qt hwc backend itself), but wouldn't you need another EGLSurface to draw upon for the external display contents?

Thanks for the feedback so far!

krnlyng:
Maybe we should compare how it's done in android, indeed the PR on qt5-qpa-hwcomposer-plugin is mirroring screen contents, but drawing different content on different screens might need a different approach

g7:
It's been a while since I last looked at how SurfaceFlinger does that, but if I remember correctly they were creating another native window per output.

This approach works for me on wlroots, see https://github.com/droidian/wlroots/pull/6/commits/20186095981af2422de8e671965bf2a7d80f7a54 ([End result](https://pbs.twimg.com/media/E_BrBeAXMAkm0aq?format=jpg&name=large)). I assume something similar can be implemented on qt5-qpa-hwcomposer.

Only blocker is being able to create more native windows in libhybris, which prompted this PR.

hybris/egl/platforms/hwcomposer/eglplatform_hwcomposer.cpp
```diff
@@ -41,22 +44,26 @@ extern "C" void hwcomposerws_Terminate(_EGLDisplay *dpy)
...
 
-	_nativewindow->common.decRef(&_nativewindow->common);
-	/* We are done with it, refcounting will delete the window when appropriate */
-	_nativewindow = NULL;
+	std::vector<HWComposerNativeWindow *>::iterator it = std::find(_nativewindows.begin(),
+		_nativewindows.end(), window);
+	if (it != _nativewindows.end()) {
+		window->common.decRef(&window->common);
+		_nativewindows.erase(it);
+	}
```
krnlyng, line 66: Should we print an error when the window isn't in the list?

mlehtima (approved):

PR #515, opened 2022-06-15
hybris: tests: Don't build vulkan test when headers aren't available.

MagneFire:
The Vulkan test will fail to compile when the relevant headers do not exist.


mlehtima:
I think the vulkan test should be built only if both WANT_WAYLAND and HAS_VULKAN_HEADER are defined. So please move the HAS_VULKAN_HEADER check and the content inside it to be within WANT_WAYLAND.

mlehtima (approved):

PR #517, opened 2022-06-15
hybris: egl: Remove Thread-Local Storage keyword to fix a segmentation fault.

MagneFire:
I'll start by mentioning that I know very little about this. I've just observed these things mentioned below and found a potential workaround. A different approach is more than welcome :smile: 

When using SDL2 based application the use of the __thread keyword caused the application to crash whenever it calls the SDL_CreateWindow() function.
Removing this keyword seems to fix this issue.
There are various references to a segmentation fault relating to the use of Thread-Local Storage, but no concrete solutions seem to be available [^1][^2].

[^1]: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81142, suggests a workaround (-mtls-dialect=gnu2) that does not appear to work.
[^2]: https://stackoverflow.com/questions/20410943/segmentation-fault-when-accessing-statically-initialized-thread-variable, mentions that it's fixed in a recent GCC, I've tested this by cross-compiling with GCC 11.2.

### DIFF
--- a/hybris/egl/egl.c
+++ b/hybris/egl/egl.c
@@ -107,7 +107,7 @@ struct ws_egl_interface hybris_egl_interface = {
 	egl_helper_get_mapping,
 };
 
-static __thread EGLint __eglHybrisError = EGL_SUCCESS;
+static EGLint __eglHybrisError = EGL_SUCCESS;
 
 void __eglHybrisSetError(EGLint error)
 {


NotKit:
Which device/SoC/Android version do you use that has this issue? It feels like there is a problem with Android/bionic libs overwriting data in TLS slots of application on host.

MagneFire:
This was tested on `sturgeon` (Huawei Watch) which has an msm8226. The Android version is 6.0.1 (`M9E42C`).

And it happens at exactly this line: https://github.com/MagneFire/gpsp/blob/master/asteroidos/hwcomposer.cpp#L41.

---

The exact commit that introduced the `__thread` was: https://github.com/libhybris/libhybris/commit/2d7b04ec8257bf95edfe06373bc67cbd81bebc2a

Initially I made some small changes to the following part to locate the segfault:
```C
EGLint androidError = _eglGetError();
if (__eglHybrisError != EGL_SUCCESS) {
    EGLint ourError = __eglHybrisError;
    __eglHybrisError = EGL_SUCCESS;
    return ourError;
}
```

You can comment out the `return ourError;` part and it will still cause it to segfault. Removing the local variable also seems to not affect the segfaulting. The moment I remove the `__eglHybrisError` parts it doesn't segfault. Which made me point to the TLS keyword. So, by trail-and-error I figured out that the keyword had something to do with it.

MagneFire:
I should probably also add that we use the `QCOM_BSP` flag, meaning that we use a patched 'HWComposer' (Available here: https://github.com/AsteroidOS/android_hardware_qcom_display/commits/marshmallow-dr1.5-release).

Do you maybe have some tips/hints on how I can debug this? Or to confirm your suspicion that the Android/bionic libs  overwrite the data in the TLS slots?

mariogrip:
@MagneFire have you tried this? https://gitlab.com/ubports/development/core/hybris-support/tls-padding

MagneFire:
@mariogrip Thanks for the hint! That indeed seems to "fix" it.

I had a look into the the SailfishOS porters log and it turns out that this is not that rare of an issue after all (@NotKit your previous help to others helped me with this issue too!). In summary there were TLS related patches missing, namely:
- https://github.com/mer-hybris/android_frameworks_native/commit/6e237efcb4ffbc58c79437ec38d4fe3e98f867cb
- https://github.com/mer-hybris/android_bionic/commit/1e92c14be252449f901ad8033a961e970e74879d

Applying those fixed the issue in a more appropriate way it also passes the TLS regression test: https://paste.pound-python.org/show/RXOUuQQKmuQ0bYSJHACy/

Thanks @NotKit and @mariogrip for helping me solve this issue!

PR #523, opened 2022-11-15
hybris: q: rename RELR sections defines to DT_ANDROID_RELR

g7:
glibc-2.36 introduces its own support for relative relocations, and does so using different constants, breaking the q linker.

Rename the Android-specific constants to DT_ANDROID_, and define (and check) the new ones if needed. Since api level 30, bionic switched to those as well [0].

[0] https://github.com/aosp-mirror/platform_bionic/commit/6663f5525dafc31de1891d7e0fa6da88c4c7dc4b

### DIFF
--- a/hybris/common/q/hybris_compat.h
+++ b/hybris/common/q/hybris_compat.h
@@ -50,10 +50,24 @@ extern "C" size_t strlcat(char *dst, const char *src, size_t size);
  * Experimental support for SHT_RELR sections. For details, see proposal
  * at https://groups.google.com/forum/#!topic/generic-abi/bX460iggiKg
  */
-#define DT_RELR 0x6fffe000
-#define DT_RELRSZ 0x6fffe001
-#define DT_RELRENT 0x6fffe003
-#define DT_RELRCOUNT 0x6fffe005
+#define DT_ANDROID_RELR 0x6fffe000
+#define DT_ANDROID_RELRSZ 0x6fffe001
+#define DT_ANDROID_RELRENT 0x6fffe003
+#define DT_ANDROID_RELRCOUNT 0x6fffe005
+
+/* Defined in glibc >= 2.36, and used in bionic since apilevel 30 */
+#ifndef DT_RELRSZ
+#define DT_RELRSZ 35
+#endif
+
+#ifndef DT_RELR
+#define DT_RELR 36
+#endif
+
+#ifndef DT_RELRENT
+#define DT_RELRENT 37
+#endif
+
 /*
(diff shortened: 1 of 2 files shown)


krnlyng (approved):
Do they have the same values?

mlehtima:
@g7 Should the changes be conditional somehow, now for Android 10 base the enum values change and don't match elf.h. On the other hand so far in Android 11 base those have never matched. Where are these enums used and how has that not been a problem so far.

g7:
The values are the same:
  * [Android 10 (DT_RELR*)](https://github.com/aosp-mirror/platform_bionic/blob/android10-gsi/libc/include/elf.h#L251)
  * [Android 11 (renamed to DT_ANDROID_RELR*)](https://github.com/aosp-mirror/platform_bionic/blob/android11-gsi/libc/include/elf.h#L537)

As far as I could see, these are only used by the linker during prelink.

This has never been a problem until glibc 2.36, where relative relocations have been implemented and the [libc headers would pull the same constants, with different values, during build](https://github.com/bminor/glibc/commit/4610b24f5e4e6d2c4b769594efa6d460943163bb) while we really want to keep the old ones around for compatibility.

Running a linker compiled with glibc 2.36 without this commit would produce something like this:

```
root@Xperia5:/home/droidian/test# test_hwcomposer 
WARNING: linker Warning: "/android/system/lib64/libhardware.so" unused DT entry: unknown OS-specific (type 0x6fffe000 arg 0x688) (ignoring)
WARNING: linker Warning: "/android/system/lib64/libhardware.so" unused DT entry: unknown OS-specific (type 0x6fffe001 arg 0x10) (ignoring)
WARNING: linker Warning: "/android/system/lib64/libhardware.so" unused DT entry: unknown OS-specific (type 0x6fffe003 arg 0x8) (ignoring)
WARNING: linker Warning: "/android/system/lib64/libcutils.so" unused DT entry: unknown OS-specific (type 0x6fffe000 arg 0x3420) (ignoring)
WARNING: linker Warning: "/android/system/lib64/libcutils.so" unused DT entry: unknown OS-specific (type 0x6fffe001 arg 0x50) (ignoring)
WARNING: linker Warning: "/android/system/lib64/libcutils.so" unused DT entry: unknown OS-specific (type 0x6fffe003 arg 0x8) (ignoring)
[snip]
Segmentation fault
root@Xperia5:/home/droidian/test# 
```

This pull request matches the bionic change and adds support for the new values as well, rather than just renaming to `DT_ANDROID_RELR*`.

PR #543, opened 2023-08-01
hooks: implement a thread-local storage workaround for Mali driver

peat-psuwit:
Hook an internal function used to obtain a TLS offset, and provide our own. This stops memory leak on Volla Phone 22, Volla Phone X23, and presumably any devices with Mali driver running Android 12 or newer.

### DIFF
--- a/hybris/common/hooks.c
+++ b/hybris/common/hooks.c
@@ -2887,6 +2887,56 @@ int _hybris_hook_android_fdsan_close_with_tag(int fd, uint64_t tag)
     return close(fd);
 }
 
+/*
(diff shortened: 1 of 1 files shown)


piggz:
Any progress on this?

piggz:
Im also shipping it for sailfish on the X23 ... would be nice to get upstreamed

piggz:
Just tagging @mlehtima @NotKit @fredldotme @krnlyng

peat-psuwit:
New version of the patch is pushed. Please have another look.

PR #544, opened 2023-08-17
Avoid uninitialised handle_ in soinfo.

dcaliste:
Since fc7bfd1b, the soinfo structure fields are uninitialised. But this is creating an issue in linker_main.cpp#823, where a temporary soinfo object is created, but its `handle_` member is not set. Then, on exit of the function, the temporary object is destroied and it mentions in Valgrind a use of an uninitialised value with the following backtrace:
```
==22274== Use of uninitialised value of size 8
==22274==    at 0x9685BB8: soinfo::~soinfo() (in /usr/lib64/libhybris/linker/q.so)
==22274==    by 0x9682B3F: android_linker_init (in /usr/lib64/libhybris/linker/q.so)
==22274==    by 0x6C363DF: ??? (in /usr/lib64/libhybris-common.so.1.0.0)
==22274==    by 0x6C36E2B: android_dlopen (in /usr/lib64/libhybris-common.so.1.0.0)
==22274==    by 0x10B0EB: __load_library (in /home/defaultuser/devel/sextant)
==22274==    by 0x10B1E3: droid_media_camera2_get_number_of_cameras (in /home/defaultuser/devel/sextant)
==22274==    by 0x10AF4B: plop() (main.cpp:15)
==22274==    by 0x10AC9F: main (main.cpp:24)
```

I don't know if it is the best way to solve the problem. Was the memset() call in soinfo constructor commented out for debugging purposes or was it harmful (because scratching initialised members) ? Is it better to put `handle_ = 0;` in the struct soinfo definition ? Should I do it for all other members ?

### DIFF
--- a/hybris/common/mm/linker.cpp
+++ b/hybris/common/mm/linker.cpp
@@ -675,6 +675,7 @@ bool soinfo::elf_lookup(SymbolName& symbol_name,
 soinfo::soinfo(const char* realpath, const struct stat* file_stat,
                off64_t file_offset, int rtld_flags) {
 
+  soname_ = nullptr;
   if (realpath != nullptr) {
     realpath_ = realpath;
   }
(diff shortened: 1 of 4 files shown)


krnlyng:
The memset was removed since it breaks stuff horribly in glibc environments. I'm not 100% sure however why that is (why it works in android to be exact).
Setting it to 0 sounds good, probably we should do that for all members. Whether you put it in the constructor or struct definition probably doesn't matter too much.

dcaliste:
> The memset was removed since it breaks stuff horribly in glibc environments.

Yeh, I understand that it can easily break every initialised members like an `std::vector<>` for instance. That's surprising that it works at all in Android, I agree.

> Setting it to 0 sounds good, probably we should do that for all members. 

Ok, I'll do it for other members where it makes sense and amend the commit and the commit message accordingly.

Thank you for the feedback.

dcaliste:
I've pushed a new version, initialising all members that don't have already a default initialisation value.

I must confess that I've no idea how to build it though :/ So I've no idea if I made a mistake. Hopefully not.

krnlyng:
lgtm, but while we're at it we should probably do this in all affected linkers.

dcaliste:
I agree. I'll do the same changes for the other linkers.

dcaliste:
Ok, I've added the same initialisations in `mm/`, `n/` and `o/` linkers.

mlehtima (approved):

PR #549, opened 2023-11-17
hybris/egl/ws: Allow hot-reloading by unloading previous WS

fredldotme:
Make libhybris respect exactly one EGL WS at a time (though also seems to be limit as to how many different EGL Displays can be given, so that support seems out of scope for this one).

This resolves an issue where eglinfo would attempt libEGL probing for various EGL platform types, allowing both the Android and Wayland WS to be used in sequence (not in parallel).

### DIFF
--- a/hybris/egl/ws.c
+++ b/hybris/egl/ws.c
@@ -27,6 +27,7 @@
 pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 
 static struct ws_module *ws = NULL;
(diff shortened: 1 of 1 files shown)


fredldotme:
After further inspection it would need some real rethinking of how to deal with that kind of layering right now (eglinfo -> glvnd -> libhybris EGL -> Android EGL), this just broke something on my device.

mlehtima (approved):

PR #555, opened 2024-04-04
tests: workaround -Wimplicit-function-declaration in system/audio.h

peat-psuwit:
system/audio.h uses strdup(), but it's exposed only with _POSIX_C_SOURCE at least 200809L. So define that.

### DIFF
--- a/hybris/tests/test_audio.c
+++ b/hybris/tests/test_audio.c
@@ -15,6 +15,9 @@
  *
  */
 
+/* Allows system/audio.h to see strdup() */
+#define _POSIX_C_SOURCE 200809L
+
 #include <android-config.h>
 #include <memory.h>
 #include <assert.h>


JamiKettunen:
Sorry for the ping @krnlyng @mlehtima but are you seeing something similar on a modern GCC compiler without this?
```
In file included from hybris/tests/test_audio.c:25:
In file included from /usr/include/android/hardware/audio.h:30:
/usr/include/android/system/audio.h:1315:12: error: call to undeclared function 'strdup'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1315 |     return strdup(param);
      |            ^
/usr/include/android/system/audio.h:1315:12: note: did you mean 'strcmp'?
/usr/include/string.h:39:5: note: 'strcmp' declared here
   39 | int strcmp (const char *, const char *);
      |     ^
```
Alternatively this also worked for me, whichever is preferred here:
```diff
--- a/hybris/tests/Makefile.am
+++ b/hybris/tests/Makefile.am
@@ -47,7 +47,7 @@ libglesv2 = $(top_builddir)/glesv2/libGLESv2.la
 endif

 test_audio_SOURCES = test_audio.c
-test_audio_CFLAGS = -std=c99\
+test_audio_CFLAGS = -std=gnu99\
       -I$(top_srcdir)/include \
       -I$(top_srcdir)/include/android
 test_audio_LDADD = \
```

PR #564, opened 2024-08-28
hybris: Don't delete displays too early, implement reference counting for eglInitialize and eglTerminate.

krnlyng:


fredldotme:
Tested on Ubuntu Touch for regressions, found none.

Test case: graphics-test-tools.eglinfo installed from the Snap Store.

mlehtima (approved):

Issue #559, opened 2024-07-14
Saving/restoring TLS pointer before and after entering bionic functions?


twaik:
Hello. I know it can be a bit weird, but can we use bionic libraries and functions without recompiling parts of firmware? 
As far as I can see the biggest problem is TLS and bionic and glibc can not share it. But what if libhybris will allocate its own TLS on thread start or first invocation on hybris function, backup original glibc TLS and reset it to newly allocated pointer before entering wrapped functions and restore it when wrapped function reaches its end?
AFAIK TLS pointer is located in some register, but I am not so sure if kernel resets it back automatically or not or how it behaves.
Thank you.

krnlyng:
I'm not sure if it's possible, someone would need to try. I also wonder if it would have a performance impact?

twaik:
Allocating and freeing memory should not have valuable impact. Saving and restoring TLS pointer should have impact similar to assigning variable so I do not think it will really matter.
You can make it a compile-time option if you want (probably should, and in this case you can avoid hooking libc functions).
But anyway it will be the only option for projects like termux, linux deploy, andronix, userland, etc. to use Android's native libraries and device drivers because most likely they run on stock firmware.

Ninlives:
Do you have any progress on this issue?

fish4terrisa-MSDSM:
Any further progress?

krnlyng:
Do you know through which mechanism they call into glibc? through a hook in hooks.c?

Those definitely each need a wrapper if we go by this approach.

I also thought about generating the code that swaps the pointers dynamically through a similar mechanism as the debug wrappers we have in hybris/common/wrappers.c etc. Maybe that's worth looking at for you too.

I'm not super convinced that this will be a performant solution, but it's nice to see someone try it so we can see some results.

twaik:
Probably it is off-topic here, but does [this](https://github.com/Linux-on-droid/lindroid-quirks/commit/86b6c113ce7a8ec86ddfd72a47e311ea4f81eb64) fixes TLS problem? If yes probably we simply could add this to `/etc/ld.so.preload`.

twaik:
Patching bionic is problematic because it will require rebuilding some components that are using TLS, like libEGL.so (which is mostly vendors' libEGL wrapper adding some extensions and ANativeWindow-related stuff).
The solution with switching TLS on every inter-libc call will cause a small overhead but will cause problems with callbacks.
[...]

PR #575, opened 2025-02-17
hybris: experimental TLS access patcher for aarch64

NotKit:
Some Android graphics drivers (noticed with libGLES_mali.so) make use of private bionic TLS slots. Since these slots use fixed offsets from the thread ID register, the access routine is inlined by the compiler and cannot be hooked by libhybris symbol overriding.

The code compiled with glibc has no notion of bionic's reserved TLS area, so these slots may be used by the executable/other libraries. While libtls-padding.so worked when TLS was relocatable, it fails when the main executable uses local-exec TLS model, where access is inlined similarly to bionic and can't be relocated by the linker.

This causes the application and GLES driver to overwrite each other's TLS data, leading to crashes.

As discussed in #559, we could try to work around this by patching TLS access dynamically in loaded libraries. When HYBRIS_TLS_PATCH=1, all libhybris linker-loaded code is analyzed for the pattern:

```
mrs x<rt>, tpidr_el0
ldr/str x<r*>, [x<rt>,#offset]
```

If found, the offset is patched to point to libhybris-allocated TLS area, avoiding conflicts between glibc and bionic-compiled code.

HYBRIS_TLS_PATCH can also be set to a colon-separated list of library filenames to limit the performance impact and avoid breaking unrelated code, e.g.:

`HYBRIS_TLS_PATCH=libGLES_mali.so`

This is pretty much initial attempt and I don't expect the current method to be reliable (it doesn't properly track register access in case there are extra instructions inserted by compiler between mrs and ldr), but workarounds the libGLES_mali.so-related crashes on affected devices.

### DIFF
--- a/hybris/common/Makefile.am
+++ b/hybris/common/Makefile.am
@@ -50,6 +50,11 @@ libhybris_common_la_SOURCES += \
 	wrapper_code_generic_arm.c
 endif
 
+if WANT_ARCH_ARM64
+libhybris_common_la_SOURCES += \
+	tls_patcher.c
+endif
+
 libhybris_common_la_CPPFLAGS = \
 	-I$(top_srcdir)/include \
 	$(ANDROID_HEADERS_CFLAGS) \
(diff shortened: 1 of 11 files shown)


NotKit:
@krnlyng I had another attempt patching in a different way with replacing mrs by trampolines, but didn't get to do any kind of performance measurements or battle-testing.

PR #580, opened 2025-05-13
compat: hwc2: allow passing nullptr buffer to setClientTarget

NotKit:
Make hwc2_compat_display_set_client_target properly handle null buffer argument by checking before dereferencing.

Passing the buffer can be skipped if the server had already cached the buffer handle in the specified slot.

### DIFF
--- a/compat/hwc2/hwc2_compatibility_layer.cpp
+++ b/compat/hwc2/hwc2_compatibility_layer.cpp
@@ -241,17 +241,20 @@ hwc2_error_t hwc2_compat_display_set_client_target(hwc2_compat_display_t* displa
                                             const int32_t acquireFenceFd,
                                             android_dataspace_t dataspace)
 {
-    android::sp<android::GraphicBuffer> target(
-        new android::GraphicBuffer(buffer->handle,
-            android::GraphicBuffer::WRAP_HANDLE,
-            buffer->width, buffer->height,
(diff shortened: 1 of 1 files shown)


krnlyng (approved):

compat/hwc2/hwc2_compatibility_layer.cpp
```diff
@@ -241,12 +241,15 @@ hwc2_error_t hwc2_compat_display_set_client_target(hwc2_compat_display_t* displa
...
 {
-    android::sp<android::GraphicBuffer> target(
-        new android::GraphicBuffer(buffer->handle,
-            android::GraphicBuffer::WRAP_HANDLE,
-            buffer->width, buffer->height,
-            buffer->format, /* layerCount */ 1,
-            buffer->usage, buffer->stride));
+    android::sp<android::GraphicBuffer> target = nullptr;
+
+    if (buffer) {
```
KonstantinRitt, line 246: perhaps worth promoting `slot` to `display->self->setClientTarget` below? ;p

also note there are some other places around where slot is hard-coded to be 0

NotKit: Thanks for noticing, updated. Regarding some other places, do you mean hwc2_compat_layer_set_buffer? In theory it's independent from setClientTarget, but worth fixing that as well.

PR #583, opened 2025-06-21
hybris: Add support for meson build system

mlehtima:


krnlyng:
looks nice, when can we drop automake :D:D

PureTryOut:
What is blocking this? I'd love a sane build system over autotools...

mlehtima:
Only needs some more testing. Some of the options are not used in Sailfish OS so those need to be verified.

PureTryOut:
Ok so tested this just now and a few things:

1. A minimum supported Meson version should be set

```
meson_options.txt:9: WARNING: Project does not target a minimum version but uses feature deprecated since '1.1.0': "boolean option" keyword argument "value" of type str. use a boolean, not a string
```

2. Various variables seem to be empty strings

```
WARNING: Required dependency was found to be an empty string. Did you mean to pass an empty array?
```

3. I compiled the autotools version without Vulkan support but that does not seem to be possible anymore with Meson.

```
Has header "vulkan/vulkan.h" : NO 

tests/meson.build:224:20: ERROR: Unknown variable "want_vulkan".
```

I could of course just build with vulkan-headers installed but it was optional before and it fails for me anyway due to issues in the Vulkan parts I have yet to fix on my Musl build (although they do not seem to be Musl related, but that's a different subject).

mlehtima:
Thanks for testing. I'll fix those issues.

PR #584, opened 2025-08-29
hybris: Fix symlinks pointing outside the source tree

KonstantinRitt:


mlehtima (approved):
Thanks for fixing this.

PR #586, opened 2025-08-31
wayland: Fix UB in destroyBuffers()

KonstantinRitt:
++it or it = erase, not both


krnlyng:
Is this really UB or leaking every second buffer? Anyways nice find.

mlehtima (approved):

PR #591, opened 2025-09-01
A couple of fixes to server_wlegl_*

KonstantinRitt:


hybris/platforms/common/server_wlegl.h
```diff
@@ -43,4 +43,7 @@ server_wlegl_create(struct wl_display *wldpy);
 void
 server_wlegl_destroy(server_wlegl *wlegl);
 
+server_wlegl *
+server_wlegl_from(struct wl_resource *);
```
mlehtima, line 47: Was there some reason to add this to header? It is only used in server_wlegl.cpp.

hybris/egl/platforms/common/eglplatformcommon.cpp
```diff
@@ -111,13 +114,11 @@ extern "C" EGLBoolean eglplatformcommon_eglQueryWaylandBufferWL(EGLDisplay dpy,
 // and decRef'ed using eglHybrisReleaseNativeBuffer
 extern "C" EGLBoolean eglplatformcommon_eglHybrisAcquireNativeBufferWL(EGLDisplay dpy, struct wl_resource *wlBuffer, EGLClientBuffer *buffer)
  {
-     if (!buffer)
```
mlehtima, line 114: Why is this check removed?

hybris/platforms/common/server_wlegl.cpp
```diff
@@ -62,10 +56,7 @@ server_wlegl_create_handle(struct wl_client *client,
 		return;
 	}
 
-	handle = server_wlegl_handle_create(id);
-	wl_array_copy(&handle->ints, ints);
-	handle->num_fds = num_fds;
-	wl_client_add_resource(client, &handle->resource);
```
mlehtima, line 68: Was this completely unnecessary or does one of the new calls in server_wlegl_handle_create do what this did?

mlehtima (changes requested):
Some small fixes and open questions.

hybris/platforms/common/server_wlegl.h
KonstantinRitt: that was simplifying testing to me, nothing more. left it here just for consistency with other headers.
should I move it to .cpp?

hybris/platforms/common/server_wlegl.cpp
```diff
@@ -146,11 +138,17 @@ server_wlegl_get_server_buffer_handle(wl_client *client, wl_resource *res, uint3
 	int r = hybris_gralloc_allocate(width, height, format, (uint32_t)usage, &_handle, (uint32_t*)&_stride);
 	if (r) {
 		HYBRIS_ERROR_LOG(SERVER_WLEGL, "failed to allocate buffer\n");
-		wl_resource_destroy(resource);
+		wl_resource_post_error(res,
+				       ANDROID_WLEGL_ERROR_BAD_VALUE,
+				       "failed to allocate buffer: possibly bad format (%d) or usage (%d)",
+				       format, usage);
```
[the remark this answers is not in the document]
KonstantinRitt: ok

hybris/egl/platforms/common/eglplatformcommon.cpp
KonstantinRitt: it is covered by `server_wlegl_buffer_from` + `!buf` below

hybris/platforms/common/server_wlegl.cpp
KonstantinRitt: has been moved to `server_wlegl_handle_create` in order hide implementation details and simplify the calling code

mlehtima: I checked the wayland code and it does indeed wl_resource_create does also do internally the same things as wl_client_add_resource.

KonstantinRitt: that's why I was calling commit "wayland: Stop using wl_resource internals in server_wlegl_handle" ;p

a forthcoming patch would move to wayland-*-core and hide protocol implementation details, thus making hybris more ABI safe

hybris/egl/platforms/common/eglplatformcommon.cpp
mlehtima: All places which use buffer in this function there is dereference of it first with *buffer so if buffer pointer is null that is bad and the check should be added back.

KonstantinRitt: ah that's about `buffer` not `wlBuffer`, got it now.

Issue #573, opened 2025-01-05
Snapped libhybris doesn't load libvulkan properly


fredldotme:
Background: I'm building a so called "provider snap" for "hybris-2404" which contains libhybris' instead of mesa's libEGL and friends.

So far OpenGL ES2 has worked properly, but there are issues with integrating libvulkan to the point where it works not even remotely.

My investigations:

Apps linking against libvulkan might crash instantly because the libhybris linker is set up with an empty environment.

I don't know what is causing the environment to be cleared at this point, needless to say all the setup instructions for working properly are ignored. It fetches the linker to use from the property socket and segfaults. I don't know how it ends up in an empty environment to begin with. 

It tries to load the right linker but fails because the path is not correct, not using the one from the environment/`getenv()` call.

fish4terrisa-MSDSM:
also got this problem, vulkan crashed immediately. Did you find any solution?

fredldotme:
Not yet. What I am unsure about is at what point `environ` is getting initialized to actual values representing the environment of the process.

It seems to me the current way of resolving wrapper symbols happens before that, which means libhybris' libvulkan implementation can only work when system-installed right now, not as a relocatable.

Issue #572, opened 2024-12-07
Vulkan support on musl libc (without GNU indirect functions)


JamiKettunen:
TITLE: Vulkan support on musl libc (without GNU indirect functions)

I've been working on https://github.com/JamiKettunen/chimera-libhybris and one of the challenges is Vulkan as https://github.com/libhybris/libhybris/blob/master/hybris/vulkan/vulkan.c makes heavy use of `gnu_indirect_function` assembly symbols which are a glibc extension.

Any ideas on how I should tackle this? I previously tried ifunc-adding patches from https://github.com/bell-sw/alpaquita-aports/tree/stream/core/musl-perf but they still seemed to result in `libvulkan.so.1`-linking programs crashing anyway until I restored them to use the `vulkan-loader` one.

Would perhaps involving the Vulkan ICD loader here be a good idea or was already ruled out before due to technical reasons or causing too much overhead (which the current ifunc-way seems to avoid after first calls to each function)?

fredldotme:
What if it had ICD entrypoints implemented in a separate .c file next to vulkan.c which we could just not compile in under those circumstances?

fredldotme:
Some insight as to what the ifunc-binding at resolve time might be doing: https://github.com/libhybris/libhybris/issues/573#issuecomment-2777584076

PR #607, opened 2026-03-06
hybris/vulkan: Do Vulkan symbol resolution without VULKAN_IDLOAD

fredldotme:
The IDLOAD approach two major drawbacks compared to regular wrapper functions:

- musl cannot make use of IDLOAD as it's a GNU libc extension
- Relocated glibc libvulkan resolves symbols before 'environ' points to a valid memory area

To counter this, create regular wrapper functions in place of IDLOAD instantiation.

The vast majority of this change was auto-generated using a conversion script, located at: https://github.com/fredldotme/libhybris-vulkan-idload-convert Two hands full of functions needed manual intervention for replacing invalid struct names with EXT or KHR equivalents.

Fixes:

- https://github.com/libhybris/libhybris/issues/572
- https://github.com/libhybris/libhybris/issues/573


fredldotme:
Closed in favor of #625

PR #610, opened 2026-04-12
egl: add EGL_EXT_platform_base to egl strings

FakeShell:
otherwise apps like fluffychat refuse to launch:

No provider of eglGetPlatformDisplayEXT found.  Requires one of:
    EGL_EXT_platform_base

hybris provides the extensions so this is safe to declare


mlehtima:
In a way the location of the changes it not the best since those are now depending on WANT_WAYLAND although the EGL_EXT_platform_base related functions are not depending on wayland. But probably wayland is always in use so it doesn't matter much.

FakeShell:
@mlehtima I can move it one block up, but I think pretty much all consumers of hybris are building with wayland so i don't think it will make any difference. let me know if you want that.

PR #612, opened 2026-04-21
egl: Implement eglCreateImage and eglDestroyImage

FakeShell:
since eglCreateImage and eglCreateImageKHR are pretty close in implementation, we can reuse eglCreateImageKHR in case the blob doesn't provide eglCreateImage (same for eglDestroyImage and eglDestroyImageKHR)

this makes flutter flatpak one.jwr.interstellar work:
No provider of eglCreateImage found.  Requires one of:
    EGL 15

### DIFF
--- a/hybris/egl/egl.c
+++ b/hybris/egl/egl.c
@@ -73,6 +73,9 @@ static EGLSurface  (*_eglGetCurrentSurface)(EGLint readdraw) = NULL;
 static EGLBoolean  (*_eglSwapBuffers)(EGLDisplay dpy, EGLSurface surface) = NULL;
 
 
+static EGLImage (*_eglCreateImage)(EGLDisplay dpy, EGLContext ctx, EGLenum target, EGLClientBuffer buffer, const EGLAttrib *attrib_list) = NULL;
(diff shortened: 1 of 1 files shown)


hybris/egl/egl.c
```diff
@@ -513,6 +516,43 @@ static EGLImageKHR _my_eglCreateImageKHR(EGLDisplay dpy, EGLContext ctx, EGLenum
...
+		image = malloc(sizeof *image);
+		image->egl_image = (EGLImageKHR) ei;
+		image->target = target;
+		image->ws_dpy = display;
+		image->ws_buffer = newbuffer;
+
+		return (EGLImage) image;
+	}
+
+	return (EGLImage) _my_eglCreateImageKHR(dpy, ctx, target, buffer, (const EGLint *) attrib_list);
```
peat-psuwit, line 547: You can't do that. `EGLint` is a different type and different size then `EGLAttrib` (`khronos_int32_t` vs. `intptr_t` (most likely 64-bit)).

FakeShell: interestingly it seems to work in my case and apps don't seem to complain. but i guess it can lead to unexpected cases too

PR #617, opened 2026-05-18
fix: define TEMP_FAILURE_RETRY for non-glibc systems

PureTryOut:
Musl libc doesn't have a definition for TEMP_FAILURE_RETRY


PureTryOut:
@mlehtima since you seem to be the most recently active on this project, could you take a look at this PR and the others I made a while ago? I need to do more fixes for Musl libc support but would like to get a few things in before I do.

mlehtima (approved):

PR #627, opened 2026-09-11
hwc2: expose display configuration API

Halhadus:
## Summary

Expose HWC2 display configuration enumeration and active configuration switching through the public compatibility API.

This adds helpers to:

- enumerate available display configurations
- retrieve configuration attributes
- switch the active configuration using `setActiveConfigWithConstraints()`

## Motivation

Consumers of the HWC2 compatibility API currently only have access to the active display configuration.

This prevents them from discovering other configurations advertised by Composer, such as alternative refresh rates, and from selecting those configurations.

The current upstream HWC2 implementation already provides `Display::setActiveConfigWithConstraints()`, including its internal fallback to the legacy configuration-switching path when appropriate. This change exposes that functionality through the compatibility API.

## Testing

The current upstream revision was compile- and link-tested as `libhwc2_compat_layer` in an Android 12 build environment.

The constraints-based public API was also tested end-to-end on a Xiaomi Poco F5 (`marble`) through the Halium 11 compatibility fallback and Droidian wlroots integration.

Composer exposes 60, 90 and 120 Hz configurations, and switching to 120 Hz succeeds.

The device runtime test exercises the public constraints API through Halium's legacy `setActiveConfig()` fallback. The current upstream internal `Display::setActiveConfigWithConstraints()` implementation itself was compile- and link-tested, but was not directly runtime-tested on this device.

## Related integration changes

- Halium/libhybris#24
- droidian/libhybris#2
- droidian/wlroots#11


compat/hwc2/hwc2_compatibility_layer.cpp
```diff
@@ -192,6 +192,50 @@ HWC2DisplayConfig* hwc2_compat_display_get_active_config(
...
+hwc2_error_t hwc2_compat_display_set_active_config(hwc2_compat_display_t* display,
+                                                    hwc2_config_t configId)
+{
+    if (!display || !display->self)
+        return HWC2_ERROR_BAD_DISPLAY;
+
+    auto configs = display->self->getConfigs();
+    for (const auto& config : configs) {
+        if (config->getId() == configId)
+            return static_cast<hwc2_error_t>(display->self->setActiveConfig(config));
```
NotKit: setActiveConfig(config) is an old API and some devices may outright ignore it (noticed on Pixel 9). Android 11 introduced setActiveConfigWithConstraints(), which implements seamless mode change on vsync/after posting a new frame. I think since we implement this anew, this function should be designed around the new API and fallback to the old call if not available.

compat/hwc2/hwc2_compatibility_layer.cpp
```diff
@@ -192,6 +192,50 @@ HWC2DisplayConfig* hwc2_compat_display_get_active_config(
...
+        return 0;
+    return display->self->getConfigs().size();
+}
+
+hwc2_error_t hwc2_compat_display_get_config(hwc2_compat_display_t* display,
+                                             size_t index,
+                                             HWC2DisplayConfig* config)
+{
+    if (!display || !display->self || !config)
+        return HWC2_ERROR_BAD_PARAMETER;
```
NotKit, line 209: can be split to return HWC2_ERROR_BAD_DISPLAY for bad display and keep HWC2_ERROR_BAD_PARAMETER for null pointer config

compat/hwc2/hwc2_compatibility_layer.cpp
Halhadus: Thanks, good point. I’ve updated the public compatibility API to use `setActiveConfigWithConstraints()` and expose the constraints/timeline data. On current upstream libhybris, the existing `Display::setActiveConfigWithConstraints()` implementation already handles the legacy `setActiveConfig()` fallback when the newer switching path cannot be used.

As a compile check, I placed this exact updated upstream revision into my Android 12 build tree and built `libhwc2_compat_layer` there. The build completed successfully, and the resulting library exports `hwc2_compat_display_set_active_config_with_constraints`.

This was only a compile/link test against an Android 12 build environment; I have not runtime-tested the new constraints-based API on a device yet.

I also checked the `halium-11.0` branch separately. Its HWC2 wrapper does not currently contain `Display::setActiveConfigWithConstraints()` or the related constraint/timeline types, and instead has the older `Display::setActiveConfig()` path. I’m looking into the appropriate backport/compatibility behavior there separately.

compat/hwc2/hwc2_compatibility_layer.cpp
Halhadus: Fixed, thanks. Invalid displays now return `HWC2_ERROR_BAD_DISPLAY`, while a null config output pointer returns `HWC2_ERROR_BAD_PARAMETER`.

